PR #4, opened 2015-06-25
Add forward-proto-header

shinji62:
Hi,

If I want to use only the port 80 as proxy (default) behind an ELB the forwarded proto should be pass, be cause some application can require this proto.

Example : 

ELB 80,443,4433--> 80 Haproxy --> 80 router  --> my app

If I query https://mydomain.com the ELB will give the https as forwarder-proto but haproxy still send http, my app is expecting https so I will send a redirect to https  .... so I will loop.


geofffranks:
Thanks!

PR #13, opened 2015-09-25
enable ssh access to containers

cromega:
We need this functionality in the release. We are not sure about adding a CF specific feature to a generic haproxy release, though. What do you think?

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -69,3 +69,14 @@ properties:
   ha_proxy.backend_port:
     description: "Listening port for Router"
     default: 80
+
+  app_ssh.servers:
+    description: "A list of hosts that run ssh proxy"
+    default: []
+  app_ssh.port:
+    description: "Enables a port for ssh access to an application container running on diego."
+    default: 2222
+
+  cc.allow_app_ssh_access:
+    default: false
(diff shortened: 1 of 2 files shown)


shinji62:
Not really convince to add CF dependency here..

cromega:
@shinji62 My thoughts exaclty. We still need port forwarding though. How about calling it something more generic? It could be *port_forwards* and the configuration value could be an array of hashes in yaml.

shinji62:
Or you can just remove cc. and use ha_proxy.

Like 
```yaml
ha_proxy.app_ssh.servers:
  description: "A list of hosts that run ssh proxy"
  default: []
ha_proxy.app_ssh.port:
  description: "Enables a port for ssh access to an application container running on diego."
  default: 2222
ha_proxy.allow_app_ssh_access:
  default: false
  description: "Allow users to change the value of the app-level allow_ssh attribute"
```


Btw I am not sure allow_app_ssh_access is require we can just check if ha_proxy.app_ssh.servers is empty.

geofffranks:
I believe this is no longer needed, due to the generic tcp proxying support from #20. If you have any issues with the proxying support provided there, please open another issue/PR.

Issue #10, opened 2015-08-24
add "https redirect" possibility


Koizumi85:
Hello,

we have a cloudfoundry deployment using the cf-haproxy-boshrelease as load balancer. We would like to redirect all users to https, so that they are not able to use http anymore.
I think everything would be okay with adding the following line in the "http-in"-frontend:

"redirect scheme https code 301 if !{ ssl_fc }"

Would it be possible to add a configuration option for this? something like "ha_proxy.always_redirect_to_https"?

Thank you very much.

longnguyen11288:
yes this would be possible, can you make PR with changes?

Koizumi85:
I am not really familiar with Pull Requests, but i will try to make one. Do i have to sign some CLA or something like this before?

longnguyen11288:
Nope!


On August 24, 2015 at 7:33:29 AM, Koizumi85 (notifications@github.com) wrote:

I am not really familiar with Pull Requests, but i will try to make one. Do i have to sign some CLA or something like this before?

ronakbanka:
@Koizumi85 Even i was thinking to put this in , If you are opening the PR i will wait or can open one if you are fine with it?

shinji62:
@Koizumi85 this PR is good training for you :)

Koizumi85:
Little question... how can i test if my change in the config-template works like expected?
I am really new with this whole cloudfoundry and template stuff and have no idea :D

The fix is a little bit different by the way. I noticed, that it would be not enough to simply enable the redirect for every domain and therefore i had the idea to add some property like "additional_http_configuration_options", where someone could provide an array with lines like "redirect scheme https code 301 if { hdr(Host) -i www.mobiti.de } !{ ssl_fc }" which are all lined up at the end of the "frontend http-in" block.
Is there anything stupid with this idea?

shinji62:
You can too options : 

1- Force https for all domain
2- If 1 is false you can give an array of domain 

If 1 is false and no domain so you have nothing where you need to force the redirect.

ronakbanka:
@shinji62 @Koizumi85 Guys how about using map feature from HAproxy.
With the host array can populate a map table and check hdr(host) against the map. Will be much faster in case of large number of hosts, as map uses tree for search instead of looking each entry in frontend .

http://blog.haproxy.com/2015/01/26/web-application-name-to-backend-mapping-in-haproxy/

shinji62:
[...]

PR #15, opened 2015-09-27
Fixed Issue #10

Koizumi85:
Following two options are possible now:
- Set property "ha_proxy.https_redirect_all" to true, redirects all requests to https
- Set property "ha_proxy.https_redirect_domains" to an array of domains, all requests to one of these domains will be redirected to https (including ALL subdomains)

This feature is using map-feature of haproxy to provide a high performance for a high number of domains in the list

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -40,6 +40,14 @@ frontend http-in
     reqadd X-Forwarded-Proto:\ http if ! xfp_exists
 <% end %>
 
+<% if p("ha_proxy.https_redirect_all") == true %>
+    redirect scheme https code 301 if !{ ssl_fc }
+<% end %>
+<% unless p("ha_proxy.https_redirect_all") == true %>
+    acl ssl_redirect hdr(host),lower,map_end(/var/vcap/jobs/haproxy/config/ssl_redirect.map,false) -m str true
+    redirect scheme https code 301 if ssl_redirect
+<% end %>
+
 <% if_p("ha_proxy.ssl_pem") do |pem| %>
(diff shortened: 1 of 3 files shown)


jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -40,6 +40,14 @@ frontend http-in
     reqadd X-Forwarded-Proto:\ http if ! xfp_exists
 <% end %>
 
+<% if p("ha_proxy.https_redirect_all") == true %>
+    redirect scheme https code 301 if !{ ssl_fc }
+<% end %>
+<% unless p("ha_proxy.https_redirect_all") == true %>
+    acl ssl_redirect hdr(host),lower,map_end(/var/vcap/jobs/haproxy/config/ssl_redirect.map,false) -m str true
```
shinji62: What's happen in the case of ssl_redirect file is empty (ha_proxy.https_redirect_domains) ?

Koizumi85: I have to confess, that i didn't test it explicitly. But as far i understand the documentation of haproxy an empty file should not cause an error and all requests should be handled without redirect.

shinji62: Just test it :)

Koizumi85: Test was successful. Could send a call without setting "https_redirect_all" and "https_redirect_domains" in my manifest

shinji62:
Ok

Koizumi85:
What do you think when you can merge my PRs and create a new final release?
Just for better planning of our update roadmap.

geofffranks:
Will do.

PR #17, opened 2016-02-05
Add capability to have multiple backends based off of URL path prefix

jasonkeene:
I needed the ability to have haproxy balance to multiple backend services based on the URL. This is the solution I came up with to try and get this merged upstream. The other option is to allow an escape hatch for users to provide their own config in their manifest but I thought this would be less invasive.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -71,7 +71,7 @@ properties:
     default:     30
 
   ha_proxy.backend_servers:
-    description: "Array of the router IPs acting as the HTTP/TCP backends (should include servers all Availability Zones being used"
+    description: "Array of the router IPs acting as the HTTP/TCP backends (should include servers all Availability Zones being used)"
     default: []
   ha_proxy.backend_port:
     description: "Listening port for Router"
@@ -79,3 +79,7 @@ properties:
   ha_proxy.compress_types:
     description: "If this property is set, gzip compression will be activated for the mime types named in this property. definition like 'text/html text/plain text/css'"
     default: ""
+
+  ha_proxy.routed_backend_servers:
+    description: "Hash of the URL prefixes -> array of the router IPs acting as the HTTP/TCP backends (should include servers all Availability Zones being used)"
+    default: {}
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -1,3 +1,5 @@
+<% require "digest"%>
+
 global
     log <%= p('ha_proxy.syslog_server') %> syslog <%= p('ha_proxy.log_level') %>
     daemon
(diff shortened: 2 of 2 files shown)


shinji62:
Can you give an example of what you try to achieve  ?

jasonkeene:
We want to route traffic to different backend applications based on URLs. For instance `/api` will balance to a set of api servers where as `/` would balance to a frontend application.

shinji62:
Any concrete final haproxy config file  ?

Of course you can remove sen. information.

After that we should be good to merge.

jasonkeene:
Here is an example manifest and rendered config we are using (names have been changed to protect the innocent):

```yaml
ha_proxy:
  disable_http: false
  ssl_ciphers: "ciphers here"
  ssl_pem: "keys here"
  backend_servers:
  - 10.0.0.1
  - 10.0.0.2
  - 10.0.0.3
  backend_port: 8080
  routed_backend_servers:
    "/v2":
      servers:
      - 10.0.0.4
      - 10.0.0.5
      - 10.0.0.6
      port: 8888
```

```
global
    log 127.0.0.1 syslog info
    daemon
    user vcap
    group vcap
    maxconn 64000
    spread-checks 4
    tune.ssl.default-dh-param 2048

defaults
    log global
    maxconn 64000
    option http-server-close
    option httplog
    option forwardfor
    option contstats
    timeout connect         5000ms
    timeout client          30000ms
    timeout server          30000ms
    timeout tunnel          3600000ms
    timeout http-keep-alive 1000ms
    timeout http-request    30000ms
    timeout queue           30000ms

frontend http-in
    mode http
    bind :80
    default_backend http-routers

    acl routed_backend_db1976 path_beg /v2
    use_backend http-routed-backend-db1976 if routed_backend_db1976

    acl xfp_exists hdr_cnt(X-Forwarded-Proto) gt 0
    reqadd X-Forwarded-Proto:\ http if ! xfp_exists

    acl ssl_redirect hdr(host),lower,map_end(/var/vcap/jobs/haproxy/config/ssl_redirect.map,false) -m str true
    redirect scheme https code 301 if ssl_redirect

frontend https-in
    mode http
    bind :443 ssl crt /var/vcap/jobs/haproxy/config/cert.pem no-sslv3 ciphers ...
    default_backend http-routers

    acl routed_backend_db1976 path_beg /v2
    use_backend http-routed-backend-db1976 if routed_backend_db1976

    reqadd X-Forwarded-Proto:\ https

backend http-routers
    mode http
    balance roundrobin

    server node0 10.0.0.1:8080 check inter 1000
    server node0 10.0.0.2:8080 check inter 1000
    server node0 10.0.0.3:8080 check inter 1000

backend http-routed-backend-db1976
    mode http
    balance roundrobin

    server node0 10.0.0.4:8888 check inter 1000
    server node0 10.0.0.5:8888 check inter 1000
    server node0 10.0.0.6:8888 check inter 1000
```

shinji62:
Ok got It, I will merge and create new final release cut a new release

PR #18, opened 2016-03-28
Added support for forcing headers on requests

geofffranks:
Added support for having HAProxy set headers on each
HTTP request, based on the `ha_proxy.headers` property.

@shinji62 Can you let me know what you think of this?

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -83,3 +83,6 @@ properties:
   ha_proxy.routed_backend_servers:
     description: "Hash of the URL prefixes -> array of the router IPs acting as the HTTP/TCP backends (should include servers all Availability Zones being used)"
     default: {}
+
+  ha_proxy.headers:
+    description: "Hash of custom headers you wish you have set on each request"
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -31,6 +31,15 @@ frontend http-in
     mode http
     bind :80
     default_backend http-routers
+<%
+if_p("ha_proxy.headers") do |headers|
+  headers.each do |header, value|
+%>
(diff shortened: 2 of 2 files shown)


jobs/haproxy/spec
```diff
@@ -83,3 +83,6 @@ properties:
   ha_proxy.routed_backend_servers:
     description: "Hash of the URL prefixes -> array of the router IPs acting as the HTTP/TCP backends (should include servers all Availability Zones being used)"
     default: {}
+
+  ha_proxy.headers:
+    description: "Hash of custom headers you wish you have set on each request"
```
shinji62: No default ?
Maybe giving a small example in the description can be useful.

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -31,6 +31,15 @@ frontend http-in
     mode http
     bind :80
     default_backend http-routers
+<%
+if_p("ha_proxy.headers") do |headers|
+  headers.each do |header, value|
+%>
+    reqadd <%= header.gsub(/(?!:\\)( )/, '\ ') %>:\ <%= value.gsub(/(?!:\\) /, '\ ') %>
```
shinji62: Please be sure that 
``` 
Any space or known delimiter must be escaped using a backslash ('\')
```

shinji62:
except my comment LGTM.

jobs/haproxy/templates/haproxy.config.erb
geofffranks: When I was testing this out, I had it auto-escaping any whitespace, but haproxy did not like backslash-literal-tab, it appears to only accept `\t` for tabs, so I dropped the auto-escaping back to only spaces, with the intent that users would escape anything more advanced. Are you aware of any other delimiters for these fields that warrant auto-escaping? Otherwise, I'll modify the description, and add an example.

shinji62:
Ok.
will merge.

PR #21, opened 2016-03-29
Added support for multiple ssl certificates

geofffranks:
Added support to provide `haproxy.ssl_pem` as an array
of certs, to enable users to serve multiple https domains through
haproxy. If `haproxy.ssl_pem` is specified as a string, a single
cert is create, the same as before. If it is an array, each element
is treated as a separate cert, which gets loaded into haproxy.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -5,12 +5,13 @@ description: "The HAProxy server can be used to terminate SSL in front of the Ro
 
 packages:
 - haproxy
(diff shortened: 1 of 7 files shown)


shinji62:
Please don't forget to rebase :)

geofffranks:
Rebased

ramonskie:
would be nice if this is also implemented in the cf-release haproxy

PR #25, opened 2016-06-29
Capability to restrict access to api endpoints.

sreenani:
Use Case:
   In the current haproxy both the cloudfoundry apps and cloudfoundry infrastructure management endpoints (like api, login, blobstore, loggregator) are exposed.

  The cloudfoundry infrastructure management endpoints are used only by cloudfondry administrator, who operates from a pre-defined set of source ip's.

   This feature is to restrict configured api endpoints to be accessible only from a fixed set of source ip's.

1) An acl is created with the src ips
2) Per api endpoint a http request deny is added, to deny the request for that api endpoint if the src ip address is not in the above acl.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -99,6 +99,15 @@ properties:
         X-Application-ID: my-custom-header
         MyCustomHeader: 3
 
+  ha_proxy.restricted_endpoints: 
+    description: "Array of api endpoints whose accessibility should be restricted from a cidr block(s) as defined the trusted_endpoints_cidrs property"
+    default: []
+    example: ["api.cloudfoundry.com", "login.cloudfoundry.com", "doppler.cloudfoundry.com"]
+  ha_proxy.trusted_endpoint_cidrs: 
+    description: "Space separated cidr blocks for restricted_endpoints"
+    default: 0.0.0.0/0
+
+
   ha_proxy.tcp:
     description: "List of mappings to perform tcp-based proxying on. See example for mapping datastructure and keys"
     default: []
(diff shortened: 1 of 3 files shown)


geofffranks:
Could this be merged into the internal-only logic, such that the internal_only_domains are only allowed from the trusted_endpoint_cidrs?

sreenani:
Modified the pull request to re-use internal-only property for trusted endpoints.
Request you to take a re-look at the pull request.

jobs/haproxy/spec
```diff
@@ -27,6 +27,9 @@ properties:
   ha_proxy.internal_only_domains:
     description: "Array of domains for internal-only apps/services (not hostnames for the apps/services)"
     default: []
+  ha_proxy.trusted_domain_cidrs: 
+    description: "Space separated trusted cidr blocks for internal_only_domains"
+    default: 0.0.0.0/0
```
geofffranks: I think the default here should be 0.0.0.0/32? As written, it seems like it's defaulting to deny `internal_only_domains` when the source is not coming from a valid IP. Right?

sreenani: 0.0.0.0/0 is allowing all source ip's by default.
http-request deny if internal !private

deny if internal domain and if source is not trusted_domain_cidrs.

geofffranks: So if an existing user of internal_only_domains has a domain set, but no trusted_domain_cidrs, the internal only domain will now become a wide-open domain, since no traffic is not in 0.0.0.0/0, so no traffic gets rejected. Switching to 0.0.0.0/32 would flip that, to put a secure default in place, such that if you have an internal_only_domain without any trusted_domain_cidrs, nothing can access the domain.

sreenani:
Made the appropriate changes to deny by default.
Request you take a look again.

geofffranks:
Awesome, thanks! Included in https://github.com/cloudfoundry-community/cf-haproxy-boshrelease/releases/tag/v8.0.4

PR #26, opened 2016-08-12
Corrected path to ssl certs for wss-in

mvdkleijn:
In various locations, the path to the ssl certs are hardcoded to `/var/vcap/jobs/haproxy/config/ssl`. Only for wss-in the path was missing the config subdir.

Making the path variable might be interesting too, but I don't have time for that right now.

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -127,7 +127,7 @@ end
 <% if p("ha_proxy.enable_4443") == true %>
 frontend wss-in
     mode http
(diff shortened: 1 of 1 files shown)


mvdkleijn:
@geofffranks I realize this is a very minor thing to do a release for, but its kinda blocking for a change I'm trying to do so I was hoping you guys could make a release for this... or can I overrule the config entry for wss-in somehow?

geofffranks:
released in 	
https://github.com/cloudfoundry-community/cf-haproxy-boshrelease/releases/tag/v8.0.5

mvdkleijn:
Thanks! Works fine. :smile: :+1:

PR #28, opened 2016-10-06
Added accept-proxy config for bind

patelpu94:
- Default accet-proxy to empty string
- Allow accept-proxy to be configured by user “accept-proxy”
- Added example settings


geofffranks:
@patelpu94 ^^

geofffranks:
thanks!

PR #34, opened 2016-11-28
enable socket level stats via port 9000

seongkki:
* This would enable a socket interface for haproxy where you can pull off network level metrics for your haproxy. 
* The UI can be reached with {YOUR_PROXY_IP}:9000/haproxy_stats with the credentials you set.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -79,10 +79,8 @@ properties:
     default: false
   ha_proxy.stats_user:
     description: "User name to authenticate haproxy stats"
-    default: admin
   ha_proxy.stats_password:
     description: "Password to authenticate haproxy stats"
-    default: admin
 
   ha_proxy.backend_servers:
     description: "Array of the router IPs acting as the HTTP/TCP backends (should include servers all Availability Zones being used)"
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -27,7 +27,7 @@ properties:
   ha_proxy.internal_only_domains:
     description: "Array of domains for internal-only apps/services (not hostnames for the apps/services)"
     default: []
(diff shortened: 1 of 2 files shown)


jobs/haproxy/spec
```diff
@@ -74,6 +74,16 @@ properties:
...
+  ha_proxy.stats_enable:
+    description: "If true, haproxy will enable a socket for stats. You can see the stats on haproxy_ip:9000/haproxy_stats"
+    default: false
+  ha_proxy.stats_user:
+    description: "User name to authenticate haproxy stats"
+    default: admin
+  ha_proxy.stats_password:
+    description: "Password to authenticate haproxy stats"
+    default: admin
+
```
geofffranks: I think I would prefer to not have default user/passwords in the specfile, to force people to provide their own.

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -24,6 +24,16 @@ defaults
     timeout http-request    <%= p("ha_proxy.request_timeout").to_i    * 1000 %>ms
     timeout queue           <%= p("ha_proxy.queue_timeout").to_i      * 1000 %>ms
 
+  <% if p("ha_proxy.stats_enable") %>
+  listen stats :9000
```
geofffranks: Should we lock this down so only certain IP ranges can access the stats, and only certain interfaces listen to relay stats? Also, what happens in the unlikely event that some app behind haproxy requires the /haproxy_stats URL?

seongkki:
@geofffranks, thanks for reviewing. Addressed your concerns and comments. Please have another look at it. Thanks.

geofffranks:
Released in v8.0.8

PR #38, opened 2016-12-07
[HAProxy] bump to 1.7.0 and add support for init-addr

krutten:


jobs/haproxy/spec
```diff
@@ -128,15 +128,24 @@ properties:
...
+          public: 8.8.8.8
   ha_proxy.dns_hold:
     description: "DNS Hold time"
     default: 10s
   ha_proxy.accept_proxy:
     description: "Allow accept proxy"
     default: false
+  ha_proxy.allow_delayed_resolution:
+    description: "Allow HAProxy to start even when a DNS may not resolve yet"
+  ha_proxy.init_addr:
```
dennisjbell: This is overly complicated.  I recommend abstracting this to take a single argument of a either an IP address or 'none', which will be appended to the end of ["last", "libc"] internally. Absent, it should default to nil.  The use cases of having full control over this attribute is extremely limited and esoteric.  I also recommend using fallback_addr to better catch the intent of this enhancement.

```
<% 
  chain = [ "last", "libc" ]
  chain << p("ha_proxy.fallback_addr") if p("ha_proxy.fallback_addr")
%>
  default-server init-addr <%= chain.join(",") %>
```

jobs/haproxy/spec
```diff
@@ -128,15 +128,24 @@ properties:
...
-        public: 8.8.8.8
+        - private: 10.0.0.2
+          public: 8.8.8.8
   ha_proxy.dns_hold:
     description: "DNS Hold time"
     default: 10s
   ha_proxy.accept_proxy:
     description: "Allow accept proxy"
     default: false
+  ha_proxy.allow_delayed_resolution:
```
dennisjbell: This is no longer applicable

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -23,6 +23,9 @@ defaults
     timeout http-keep-alive <%= p("ha_proxy.keepalive_timeout").to_i  * 1000 %>ms
     timeout http-request    <%= p("ha_proxy.request_timeout").to_i    * 1000 %>ms
     timeout queue           <%= p("ha_proxy.queue_timeout").to_i      * 1000 %>ms
+    <% if p("ha_proxy.init_addr") %>
```
dennisjbell: See above review

jobs/haproxy/spec
```diff
@@ -128,15 +128,24 @@ properties:
...
           ssl: true        # optional - enables ssl, and uses the `ha_proxy.ssl_pem` provided key
+
   ha_proxy.resolvers:
     description: "List of DNS servers"
     example:
       resolvers:
-        private: 10.0.0.2
-        public: 8.8.8.8
+        - private: 10.0.0.2
+          public: 8.8.8.8
```
dennisjbell: Is this suppose to be an array containing a single two-item hash, or an array of two single-item hashes?

krutten: When testing, the code requires the array of keypairs

geofffranks:
@joek I know you have skin in the HAProxy resolvers functionality, do you want to weigh in on whether this change will negatively affect your use case at all?

geofffranks:
Also curious how setting init_addr globally as a default server works when you have backends pointed to different places via DNS + the tcp backends, where you can have a different backend set per tcp port haproxy listens on. It seems like setting them all to the same default backend IP until resolvers can be contacted is a bad idea there.

jobs/haproxy/templates/haproxy.config.erb
krutten: The init_addr is independent of the resolver and will deploy successfully with and without the resolver section.  Previously the deploy would fail because with Vault the domain name disappears when bosh restarted the Vault process and HAProxy error's out when it tried to restart.

The resolvers section was needed to make it useful.  I had to add

```
        resolvers:
          - consul: 127.0.0.1
```

so it would query consul in this case.  Without the resolver section, it would use the default behaviour.

I coded this behaviour instead of a "true/false" for my use case because there are many variations that make sense and exposing the raw line makes using this much simpler.

jhunt:
@krutten any movement on the changes requested by @dennisjbell ?

lcacciagioni:
@krutten I think that when you use haproxy + consul maybe is better to combine it with [consul-template](https://github.com/hashicorp/consul-template) and use a proper reload command / signal to make haproxy read the configuration again. You can see an example on how to start [here](https://github.com/hashicorp/consul-template/blob/master/examples/haproxy.md). Please let me know if you need help.

PR #39, opened 2016-12-13
Keepalived support

axelfauvel:
Adding support for [keepalived](http://www.keepalived.org/documentation.html) to enable high availability in an haproxy deployment, leveraging the https://en.wikipedia.org/wiki/Virtual_Router_Redundancy_Protocol more formally [RFC 5798](https://tools.ietf.org/html/rfc5798). See [keep-alived man page](https://linux.die.net/man/5/keepalived.conf) for more precision over capabilities of keep-alived as well as the [keep-alived user manual](http://www.keepalived.org/pdf/UserGuide.pdf)

This enables declaring a virtual IP (``keepalived.vip``) that will automatically fail over between the multiple haproxy VMs: the master will initially be the bosh vm for haproxy job instance 0. The default IP addresses assigned by bosh to vms on eth0 are used within the VRRP protocol.

Notes:
* this PR needs to create a new blob in the cf-community S3 bucket. In order to avoid clean ups if the PR is not merged, we did not yet upload the blob. The corresponding sources are at http://www.keepalived.org/software/keepalived-1.2.24.tar.gz. Let us know if you'd prefer instead us to upload the blob on the Cf community s3 blob and reference it into the PR.
* limitations and future work documented in the  docs/keepalived.md file
* the template jobs.yml was not modified to systematically include the keepalived into the haproxy jobs (as this would trigger systematic enabling od keepalived). 
* The keepalived job may also be collocated with other bosh releases such as the mysql release switchboard (see related discussion https://github.com/cloudfoundry/cf-mysql-release/issues/82#issuecomment-150704953 ). It may make sense to extract the keepalived package and jobs into its own release in the future.

### DIFF
--- a/config/blobs.yml
+++ b/config/blobs.yml
@@ -19,3 +19,8 @@ haproxy/socat-1.7.3.1.tar.gz:
   object_id: bcc504b7-e970-4469-ad93-d030b082e14c
   sha: a6f1d8ab3e85f565dbe172f33a9be6708dd52ffb
   size: 606049
+keepalived/keepalived-1.2.24.tar.gz:
+  object_id: aa936bbe-fe82-49cc-82c8-99d867e5b4bb
+  sha: !binary |-
+    YTY5YjJjNDA2MjdhOWJkNjk2OThhNTdlODFhOGM4Yzk3ODI2MDI1ZA==
+  size: 601873
--- a/docs/keepalived.md
+++ b/docs/keepalived.md
@@ -1,0 +1,103 @@
+# Purpose of keepalived implementation
+
+Adding support for [keepalived](http://www.keepalived.org/documentation.html) to enable high availability in an haproxy deployment, leveraging the https://en.wikipedia.org/wiki/Virtual_Router_Redundancy_Protocol more formally [RFC 5798](https://tools.ietf.org/html/rfc5798). See [keep-alived man page](https://linux.die.net/man/5/keepalived.conf) for more precision over capabilities of keep-alived as well as the [keep-alived user manual](http://www.keepalived.org/pdf/UserGuide.pdf)
+
+This enables declaring an virtual IP (``keepalived.vip``) that will automatically fail over between the multiple haproxy VMs: the master will initially be the bosh vm for haproxy job instance 0. The default IP addresses assigned by bosh to vms on eth0 are used within the VRRP protocol.
+
+Prereqs:
+ * The haproxy VMs must be within the same broadcast domain, i.e. receive multicast traffic sent to the 224.0.0.18 broadcast and IP protocol number 112.
+* The clients using this VIP must be within the [same broadcast domain](https://en.wikipedia.org/wiki/Broadcast_domain) as the haproxy vms and accepting ARP gratuitious. 
+
+
+# This feature has been successfully tested on the following IAAS :
+* Cloudstack w/ XenServer
+
+
+# Limitations and future enhancements
+* logs collection and monitoring/alerting : keepalived logs are sent to syslog and can t be retrieved using `bosh logs` you have to tail /var/log/syslog to get info
+* Health check period is hardcoded to 2s : we will add parameter for this
+* mcast_src_ip @IP is 224.0.0.18 : we will add parameter for this
+* Not yet email notification : we will add parameter for this
+* Hardcoded VRRP advertisement to 1 S (advert_int) triggering a new VRRP election and fail over. Not yet drain script handling to prevent downtime while bosh upgrades.
+* For the moment, KeepAlived is configured to use broadcast for network communication between nodes. Future versions will be able to use unicast to expose a VIP or control a distinct SDN system such as an AWS ElasticIP (through custom VRRP failover notification scripts)
+
+
+# testing
+## First verification
+* after setting up keepalived.vip parameter, connect to the instance with index 0 of your AZ. BOSH will configure this one as master
+* run `sudo ip a`
+* you should see the VIP (in example above, VIP is set as 10.234.250.201)
+
+```
+2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
+    link/ether 06:c9:f6:00:0a:38 brd ff:ff:ff:ff:ff:ff
+    inet 10.234.250.199/26 brd 10.234.250.255 scope global eth0
+       valid_lft forever preferred_lft forever
+    inet 10.234.250.201/32 scope global eth0
+       valid_lft forever preferred_lft forever
+```
+* The VIP is up, you can perform further testing and access your backend services using the VIP
+
+## Failover scenario
+* Let s stop haproxy on first node by running `monit stop haproxy`
+* Let s run `ip a` on first node
+```
+2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
+    link/ether 06:c9:f6:00:0a:38 brd ff:ff:ff:ff:ff:ff
+    inet 10.234.250.199/26 brd 10.234.250.255 scope global eth0
+       valid_lft forever preferred_lft forever
+```
+* no more VIP, let s look at our second node
+```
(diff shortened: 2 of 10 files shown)


config/final.yml
```diff
@@ -4,4 +4,9 @@ blobstore:
   options:
     bucket_name: cf-haproxy-boshrelease
 
+#blobstore:
+#  provider: local
+#  options:
+#    blobstore_path: /tmp/haproxy-blobs
+
```
geofffranks: This section should probably be removed before the final merge

docs/keepalived.md
```diff
@@ -0,0 +1,96 @@
+# Purpose of keepalived implementation
+
+Adding support for [keepalive](http://www.keepalived.org/documentation.html) to enable high availability in an haproxy deployment, leveraging the https://en.wikipedia.org/wiki/Virtual_Router_Redundancy_Protocol more formally [RFC 5798](https://tools.ietf.org/html/rfc5798). See [keep-alived man page](https://linux.die.net/man/5/keepalived.conf) for more precision over capabilities of keep-alived as well as the [keep-alived user manual](http://www.keepalived.org/pdf/UserGuide.pdf)
+
+This enables declaring an virtual IP (``keepalived.vip``) that will automatically fail over between the multiple haproxy VMs: the master will initially be the bosh vm for haproxy job instance 0. The default IP addresses assigned by bosh to vms on eth0 are used within the VRRP protocol.
+
+Prereqs:
+ * The haproxy VMs must be within the same broadcast domain, i.e. receive multicast traffic sent to the 224.0.0.18 broadcast and IP protocol number 112.
+* The clients using this VIP must be within the [same broadcast domain](https://en.wikipedia.org/wiki/Broadcast_domain) as the haproxy vms and accepting ARP gratuitious
+
```
geofffranks: Another potential pre-req - does the IaaS need to support traffic from multiple IPs originating from the same host (see the `source_dest_check` cloud property for AWS: https://bosh.io/docs/aws-cpi.html)?

Not sure how that applies to other infrastructures and their networking layers. Some might be fine, some might have issues. Not necessarily a bad thing, but something to make people aware of?

Also, do users need to pick a VIP for haproxy, and reserve that in BOSH networking, so BOSH doesn't attempt to dole out the address to VMs with dynamic IP addressing?

svrc-pivotal: yeah this won't work in aws, gcp, or azure; only vsphere most likely, as it requires broadcast.  there is a way to make it work with those clouds but you'd need to have forked scripts for each that use their CLI to rebind the public/elastic IP on failover, e.g. 

https://blog.logentries.com/2014/12/keepalived-and-haproxy-in-aws-an-exploratory-guide/

of course rebinds can take a couple of minutes so not quite "HA" but better than nothing ;)

geofffranks:
I believe it can work in AWS, if the VMs have the source-dest-check disabled (which was a feature added to the AWS CPI back in August)

svrc-pivotal:
the VRRP protocol itself won't work on AWS though because it uses both IP multicast for heartbeats and gratuitous ARP to manage the failover.  

you need to use unicast with a custom script.

geofffranks:
Ah, fair enough.  Thanks!

config/final.yml
axelfauvel: fixed in 1eab6d86d6e10d32aa9277ee6c51c224178c2a92

gberche-orange:
Thanks @geofffranks and @svrc-pivotal for your feedback, sorry for late response.

> Also, it looks like there are merge conflicts that need resolution.

Sorry about that, we're fixing that.

> How much has this been tested in live environments? 

We've successfully tested on CloudStack with XenServer hypervisor (TBC [advancednetworking](http://docs.cloudstack.apache.org/projects/cloudstack-installation/en/4.6/network_setup.html), we yet have to study whether this could work within a cloudstack VPC). 

We yet have to test it on our openstack environment (cloudwatt.com using KVM and opencontrail), as well as vmware vcloud (TBC with vmnet? networking). 

We have not yet tested either on bosh-lite.

> Is there a way to not use keepalived? 
> If we can't turn off keepalived, and an IaaS can't support multiple IPs from the same VM on a single interface, we might run into problems. Maybe it should be split into a second job to support that case?

We originally considered adding a short-term hack in ctl/monit templates to not start keepalived unless the ``keepalived.vip`` property is set, but this would be thrown away soon.

So, instead we'll indeed split into a distinct keepalived job (possibly next week or early january). This in sync with a potential future extraction in a distinct release (enabling cleaner colocation into other releases such as cf-mysql-release, see related https://github.com/cloudfoundry/cf-mysql-release/issues/82#issuecomment-150704953)

> How does it behave in a single-node (bosh-lite) environment?

Not sure yet, tests and inputs from the community would be useful if we don't manage to prioritize this. What would be the use-cases for VRRP on bosh-lite ? 

> I've long wanted this kind of feature here but doing it right multi-cloud requires a fair bit of engineering/testing

Agreed. Let's add per Iaas HOWTOs in the documentation, and capture the valuable information exchanged in this discussion for others to build on. The support for other Iaas with unicast properties and specific notify_master scripts would be good future additions.

> This PR needs to create a new blob in the cf-community S3 bucket. In order to avoid clean ups if the PR is not merged, we did not yet upload the blob. The corresponding sources are at http://www.keepalived.org/software/keepalived-1.2.24.tar.gz. Let us know if you'd prefer instead us to upload the blob on the Cf community s3 blob and reference it into the PR.

We'll push source blobs in the cf-community S3 bucket with @poblin-orange AWS access key. Let us know if you see an issue there. This will make testing easier until the PR gets merged.

Thanks again,

Axel & Guillaume.

geofffranks:
Feel free to push the blobs. I'm fine with this being added as a works-on-some-IaaSes-only thing, so long as we document that in the README as you suggested, and provide a way to disable it for environments it can't be supported (splitting it out into its own job should be sufficient for that). 

My concerns with single-node clusters/bosh-lite go away if it gets factored out into its own job as well, since it can then be disabled for those environments. I was mainly concerned with the ability to easily deploy + test other haproxy changes in bosh-lite, as well as our CI environment.

Thanks again for this feature!

shinji62:
Please squash your commit ! Impossible to follow.

axelfauvel:
commit squashed, blobs pushed and tested with latest master version

geofffranks:
LGTM, any additional thoughts @shinji62 ?

config/blobs.yml
```diff
@@ -19,3 +19,8 @@ haproxy/socat-1.7.3.1.tar.gz:
   object_id: bcc504b7-e970-4469-ad93-d030b082e14c
   sha: a6f1d8ab3e85f565dbe172f33a9be6708dd52ffb
   size: 606049
+keepalived/keepalived-1.2.24.tar.gz:
+  object_id: aa936bbe-fe82-49cc-82c8-99d867e5b4bb
+  sha: !binary |-
```
axelfauvel: Works well on my side. Are you afraid of something ?

gberche-orange:
@shinji62 @geofffranks anymore pending concerns before merging this PR ?

shinji62:
@gberche-orange Should be ok, @geofffranks Could merge this PR.

geofffranks:
Yep, sorry. Thought there was more on the sha1 discussion. Will merge + release this morning

gberche-orange:
thanks @geofffranks for the merge!

cppforlife:
might be a bit late, but it feels like it should just be a separate generic keepalived release, so that it could be colocated with anything.

gberche-orange:
@cppforlife , yes we indeed plan to extract keepalived support into a distinct release as mentionned in https://github.com/cloudfoundry-community/haproxy-boshrelease/pull/39#issuecomment-267558662 as to enable easier reuse in other jobs that require HA such as:
- container networking's  VTEP gateway 
- services such as mysql proxy job

We're wondering whether the [networking-release](https://github.com/cloudfoundry/networking-release) could be a good recipient for the keep-alived job ?

If so, would a PR on the networking-release likely to be accepted ?

geofffranks:
Any reason to not create a standalone keepalived-boshrelease?

axelfauvel:
not that i can think of

PR #44, opened 2017-01-07
Enable SSL backends with CA cert verification

svrc-pivotal:
For users that really want hop by hop SSL termination.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -13,6 +13,7 @@ templates:
   monit_debugger:            bin/monit_debugger
   certs.ttar.erb:            config/certs.ttar
   ssl_redirect.map.erb:      config/ssl_redirect.map
+  backend-ca-certs.erb:      config/backend-ca-certs.pem
   helpers/ctl_setup.sh:      helpers/ctl_setup.sh
   helpers/ctl_utils.sh:      helpers/ctl_utils.sh
   properties.sh.erb:    data/properties.sh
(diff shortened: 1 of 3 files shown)


geofffranks:
I like this as a feature add, but have a couple requests:

1. Could this be simplified a little, to one property for determining whether it verifies?
2. Can you move the logic for generating the ssl directives up where it can be easier to read, vs all on one line?
3. I would like to see  support for still verifying SSL certificates, but not providing a ca_file (seems like right now, you either have it on, and provide a ca_file, or it's disabled)
4. It seems like verifyhost should only be added if ssl verification is enabled, but this looks like it will add it regardless, if the property is set.
5. Does this support need to be extended to the other backends in the haproxy config (http-routed-backend, and tcp-backends)?

svrc-pivotal:
Hi Geoff

I have a commit to simplify this ;  some comments
1. will do, i'll change the setting from boolean to an enumeration, one of "verify, noverify, anything else implies off".
2. will do but haproxy requires a single line so I'll have to use ERB trim mode creatively
3. ca file is mandatory for ssl cert verification; I've clarified the description that it's optional and only needed for verify mode. 
4.  fixed
5.  Yeah, ideally we would have an ssl verify setting per backend.    I'll include this.  

One final note is the assumption of a single backend_ca_file for SSL verification is that all the backends would have the same CA chain.    I think this is reasonable until someone comes up with a case where you're proxying different private CA backends (I'd think it would be rare).

PR #45, opened 2017-01-07
OpenSSL 1.0.2j support for faster TLS on Intel chipsets 

svrc-pivotal:
OpenSSL 1.0.2j support for faster TLS on Intel chipsets with AVX/AVX2 and AES-NI extensions, along with FIPS support (see https://software.intel.com/en-us/articles/improving-openssl-performance)

I know this would put this release on the OpenSSL upgrade treadmill so I understand if you're not interested in folding this one in, just noting that I'm seeing a number of companies wanting to use a recent OpenSSL with their HAproxy in front of CF for performance, I've got a private bucket final bosh release here:  https://github.com/svrc-pivotal/haproxy-boshrelease/tree/svrc  

blobs are here
https://www.openssl.org/source/openssl-fips-2.0.14.tar.gz
https://www.openssl.org/source/openssl-1.0.2j.tar.gz


shinji62:
Hi,

Not sure is a good idea to put more security concern here by adding to support OpenSSL.


Thanks.

svrc-pivotal:
I agree it's a tradeoff; ultimately we're talking 2x or more performance gains though vs. OpenSSL 1.0.1f on the stemcell.  Shrug, if others are uncomfortable I'll close this and keep talking to Pivotal R&D .... as there are a pile of features in this community HAProxy release that should be in cf-release or PCF.

geofffranks:
What're the chances of including this in the base stemcells, so that other software can magically make use of the benefits? I think that would be my preferred route if its feasible.

svrc-pivotal:
Haha, ha, ugh.  Probably when we move to Ubuntu Xenial 16.

svrc-pivotal:
I'll get back to you on that one later this week

jhunt:
@svrc-pivotal any update?

svrc-pivotal:
The usual issue of not wanting to have to keep this up to date outside of the usual Ubuntu 14.04 LTS fix process.  I'm going to keep poking at this internally as I will probably kill myself if we're stuck with OpenSSL 1.0.1 + fixes for the next couple of years.

geofffranks:
@svrc-pivotal should we close this out, or keep trying for something here?

svrc-pivotal:
@geofffranks Now that Pivotal is supporting this release in PCF it may become a thing.   I've asked the PM.

geofffranks:
Closing due to inactivity and the prevalence of new stemcells with new ssl support. If that's an issue, feel free to re-open

PR #46, opened 2017-01-24
Convert templates + make_manifest to BOSH 2.0 manifest

geofffranks:


geofffranks:
@shinji62 what do you think of this approach?

templates/make_manifest
```diff
@@ -7,18 +7,18 @@ STEMCELL_OS=${STEMCELL_OS:-ubuntu}
...
-    [ "$infrastructure" != "openstack" ] && \
-    [ "$infrastructure" != "warden" ] || \
-    [ -z "$servers" ]; then
-  echo $infrastructure
-  echo "usage: ./make_manifest <aws-ec2|warden|openstack> <backend_server1>,[<backend_server2>,...,<backend_serverN>]"
+if [[ "$infrastructure" != "aws-ec2" ]] && \
+    [[ "$infrastructure" != "openstack" ]] && \
+    [[ "$infrastructure" != "warden" ]] || \
+    [[ -z "$servers" ]]; then
+  echo >&2 "usage: ./make_manifest <aws-ec2|warden|openstack> <backend_server1>,[<backend_server2>,...,<backend_serverN>]"
```
jhunt: Does the script even use `$infrastructure` any more, given the use of cloud-config?

templates/make_manifest
```diff
@@ -64,21 +64,37 @@ release=$templates/..
...
+if [[ $(bosh cloud-config | spruce json | jq -r '.networks[] | select(.name == "haproxy") | .name') != "haproxy" ]]; then
+	echo >&2 "Your cloud-config does not have the 'haproxy' network in it. See templates/cloud-config-warden.yml for an example."
+	exit 1
+fi
+if [[ $(bosh cloud-config | spruce json | jq -r '.vm_types[] | select(.name == "haproxy") | .name') != "haproxy" ]]; then
+	echo >&2 "Your cloud-config does not have the 'haproxy' vm_type in it. See templates/cloud-config-warden.yml for an example."
+	exit 1
+fi
+
+bosh cloud-config | spruce merge --cherry-pick networks > $tmpdir/cc-networks.yml
```
jhunt: Since this is a relatively new feature of spruce, should we put a version check in for a little bit?  Alternatively, what happens if you use older spruces with this setup?

templates/make_manifest
geofffranks: It's currently used to figure out what stemcell to auto-upload, if none are present on the director, but I imagine that could be worked around or removed entirely.

templates/make_manifest
geofffranks: Yeah, version check added

Issue #55, opened 2017-04-26
Add links support


rkoster:
It would be nice if the haproxy-boshrelease would consume links [just like the haproxy from cf-release](https://github.com/cloudfoundry/cf-release/blob/master/jobs/haproxy/spec#L18-L25).

lcacciagioni:
@rkoster Is the ssh port hardcoded in [here](https://github.com/cloudfoundry/cf-release/blob/master/jobs/haproxy/templates/haproxy.conf.erb#L92)? Is that right?

geofffranks:
I'd like to keep this release cf-agnostic, since it's not always deployed in front of it. The equivalent of the router link should probably be something like `http_backend`, and consume the port to send backend requests to as well.

For the ssh-proxy, that typically goes in the tcp-backends section, though I'm not aware of any feature to have an arbitrary array of links being consumed to set up multiple tcp-backends. You could always make use of the `ssh-proxy.service.cf.internal` hostname as the backend server, or potentially we could add an arbitrary number (5? 10?) of tcp-backend links that get processed, though the links would likely need to include some sort of way to map to configuration saying what the frontend port is,  and ssl settings for the frontend/backend

geofffranks:
Closed via #56

PR #56, opened 2017-05-03
Links 

rkoster:
This PR addresses: https://github.com/cloudfoundry-incubator/haproxy-boshrelease/issues/55.
I would have liked this PR to be more generic, however I was unable to find a way to consume links with conflicting link types.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -16,7 +16,16 @@ templates:
   backend-ca-certs.erb:      config/backend-ca-certs.pem
   helpers/ctl_setup.sh:      helpers/ctl_setup.sh
   helpers/ctl_utils.sh:      helpers/ctl_utils.sh
-  properties.sh.erb:    data/properties.sh
+  properties.sh.erb:         data/properties.sh
+
+consumes:
+  - name: http_backend
+    type: http_backend
+    optional: true
+
+  - name: tcp_backend
+    type: tcp_backend
+    optional: true
 
 properties:
   ha_proxy.syslog_server:
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -24,7 +24,7 @@ consumes:
     optional: true
 
   - name: tcp_backend
-    type: tcp_backend
+    type: ssh_proxy
     optional: true
 
 properties:
@@ -174,6 +174,8 @@ properties:
                                  # Setting `verify` requires `ha_proxy.backend_ca_file` key to be set.
           backend_verifyhost: example.com # optional - hostname to verify in the x509 certificate subject for SSL-enabled backend servers
                                           # only used if backend_ssl: `verify` is set
+  ha_proxy.tcp_link_port:
+    description: "Port haproxy should listen on when using the tcp_backend link"
   ha_proxy.resolvers:
     description: "List of DNS servers"
     example:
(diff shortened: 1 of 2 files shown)


jobs/haproxy/spec
```diff
@@ -165,6 +174,8 @@ properties:
                                  # Setting `verify` requires `ha_proxy.backend_ca_file` key to be set.
           backend_verifyhost: example.com # optional - hostname to verify in the x509 certificate subject for SSL-enabled backend servers
                                           # only used if backend_ssl: `verify` is set
+  ha_proxy.tcp_link_port:
+    description: "Port haproxy should listen on when using the tcp_backend link"
```
geofffranks: Shouldn't this be something we pull from the link itself?

rkoster:
I would love to, however at this point the diego release is not exposing this property:
https://github.com/cloudfoundry/diego-release/blob/develop/jobs/ssh_proxy/spec#L13-L15 <https://github.com/cloudfoundry/diego-release/blob/develop/jobs/ssh_proxy/spec#L13-L15>
I have tried specifying link properties (http://bosh.io/docs/links-manual.html <http://bosh.io/docs/links-manual.html>) but this is not allowed when also doing normal linking..

geofffranks:
Yeah, I hadn't yet gotten to the part where you try to use the link then fail-over. It's all good.

Though looking at the bosh.io <http://bosh.io/> docs, the way you're calling p() isn't mentioned (to provide a default if not present). I assume it works, since you were testing with diego not providing the property, right?

rkoster:
@geofffranks [here](https://github.com/cloudfoundry/bosh/blob/master/src/bosh-template/lib/bosh/template/evaluation_context.rb#L65-L70) you can find defaults for `p()`.

geofffranks:
Thanks!

geofffranks:
Soon as release notes get through CI

rkoster:
nice!

ematpl:
Hey, @geofffranks, I'm a little surprised that you accepted this PR with the `tcp_backend` link as type `ssh_proxy` instead of as a more generic identifier, given that it's still essentially specific to the [`ssh_proxy` job in diego-release](https://github.com/cloudfoundry/diego-release/blob/v1.22.0/jobs/ssh_proxy/spec#L13-L15). I'd be open to having that job in diego-release provide a separate link type (say, `tcp-backend`) if it allows the HAProxy release to be more generic. Along those lines, it also seems like it would be beneficial to supply the backend port in the link: the template is looking for it at the reasonable location of `port`, but unfortunately the `ssh_proxy` job embeds that port in its [`diego.ssh_proxy.listen_addr` property](https://github.com/cloudfoundry/diego-release/blob/v1.22.0/jobs/ssh_proxy/spec#L25-L27), and BOSH links don't yet offer a mechanism to rename or otherwise manipulate properties in the providing job to different properties in the link. That said, I'm starting to consider how to de-namespace and simplify the diego-release job properties now that BOSH recommends scoping the properties in a manifest to each job in an instance group, and it would be reasonable there to extract that port into a separate property (say, `port` or `server_port`) that would be more natural to include in the links that job provides.  

Also pinging @cppforlife and @Amit-PivotalLabs for any suggestions they may have on link design and/or potential forthcoming link capabilities.

Thanks again,
Eric

PR #58, opened 2017-05-29
Enable ha proxy to have bind ip

datianshi:
I would like to build VM with multiple ethernet. But did not find option to bind a specified IP. So provided this property


geofffranks:
@datianshi can you add the bind IP support to the 4443 frontend + the frontends for the tcp mode?

shinji62:
Agree with @geofffranks `all or nothing` :)

datianshi:
@geofffranks  Added the binding IP to 4443. Also squashed with previous commit.
Thanks

geofffranks:
@datianshi how about the `bind` directive for `tcp-frontend-*`?

datianshi:
sorry missed the tcp mode... already changed the commit and squashed @geofffranks

geofffranks:
👍 thanks!

PR #59, opened 2017-06-23
Adding support for the tcp backend_port setting in non-statically mapped backends

freynca:
Signed-off-by: Josh Winters <jwinters@pivotal.io>

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -280,7 +280,7 @@ if_link("tcp_backend") do |tcp_backend|
     "name" => tcp_backend.instances.first.name || "link",
     "backend_servers" => tcp_backend.instances.map(&:address),
     "port" => tcp_backend.p("port", p("ha_proxy.tcp_link_port")),
-    "backend_port" => p("ha_proxy.backend_port")
+    "backend_port" => tcp_backend.p("backend_port", p("ha_proxy.tcp_link_port"))
   }
 end %>
 <% tcp.each do |tcp_proxy| %>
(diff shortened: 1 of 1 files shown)


geofffranks:
Seems like this would set the backend_port for tcp backends using links to the same port as the http backends, rather than the inbound port, would it not?

So, for the case of ssh_proxy on port 2222, the link provides 2222, and which haproxy would listen on, but then send traffic to 80 (default `http.backend_port` value) due to https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/master/jobs/haproxy/templates/haproxy.config.erb#L302-L305

freynca:
@geofffranks The purpose of the change is to allow port remapping for TCP connections when links are used. Currently, the only way to populate the `backend_port` setting in TCP mode is with static mapping (https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/master/jobs/haproxy/spec#L177). Is there a better way of doing remapping/forwarding?

geofffranks:
Currently, the behavior for tcp frontend/backends is as follows:

If a link is detected, add an item to the list of TCP backends using the IPs of the link for backend hosts, and its provided port for a front/backend port. When iterating over all tcp backends defined, prefer the `backend_port` property of the individual backend, if specified. Otherwise use the `port` for both frontend and backend.

If you have a link that does not provide a port, the `ha_proxy.tcp_link_port` value will be used, and this is the port used for front/backend configuration. 

Are you trying to configure something with a link that does not provide a port, and you need differing frontend and backend ports?

freynca:
The link may provide a [frontend] `port`, but it does not provide a `backend_port`, which is what we're looking to configure. Our attempt at adding `backend_port` as a custom property to the link in the underlying release did not work. We're using HAproxy as a TCP passthrough to a back-end service, with differing frontend/backend ports.

geofffranks:
The port property from the link doesn't advertise the port on which the IPs included in the link are listening on?

freynca:
Fixed - see https://github.com/cloudfoundry-incubator/haproxy-boshrelease/pull/59/commits/00ad2faf6df1053e7602d712792784df506e3d95

PR #60, opened 2017-06-28
Add hostname from Host header to http log

jgf:
This change adds the first 20 characters of the 'Host' header to the log. It helps when you try to find out to which applications the requests are sent.

A shortened example of a line in the logfile. The captured host name is highlighted:

Jun 28 14:26:15 localhost haproxy[4812]: 52.52.52.52:1337 [28/Jun/2017:14:25:45.357] https-in~ http-routers/node0 30008/0/0/7/30015 200 1442 - - ---- 3/3/2/2/0 0/0 **{api.cf.cf-hcp-wot.aw}** "GET /internal/bulk/apps?batch_size=500&format=fingerprint&token={} HTTP/1.1"

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -77,6 +77,7 @@ end
 frontend http-in
     mode http
     bind <%= p("ha_proxy.binding_ip") %>:80
+    capture request header Host len 256
     default_backend http-routers
 <%
 if_p("ha_proxy.headers") do |headers|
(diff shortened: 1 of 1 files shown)


geofffranks:
Any reason for only limiting it to 20 characters vs 256?

jgf:
Good point. In our case we found out that 20 chars is long enough to find out which host is the target and we also didn't want to clutter the output too much. However I see the benefit of more (256) characters. It is  likely the better choice.

jgf:
We discussed this issue and came to the conclusion that capturing the full 256 chars is a good idea. Thanks for the input. I changed this pull request accordingly.

PR #61, opened 2017-06-28
Add option to set number of haproxy threads per VM

jgf:
This optional parameter allows to set the number of haproxy threads started for each VM. This option helps to improve the performance - especially on multicore VMs.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -111,10 +111,10 @@ properties:
     default:     30
 
   ha_proxy.stats_enable:
-    description: "If true, haproxy will enable a socket for stats. You can see the stats on haproxy_ip:9000/haproxy_stats"
+    description: "If true, haproxy will enable a socket for stats. You can see the stats on `haproxy_ip:9000/haproxy_stats`. If multithreading is enabled (`ha_proxy.threads > 1`) haproxy will create a separate socket and stat page for each thread. Each stat page is reachable on a different port ranging from `9000` to `9000 + ha_proxy.threads - 1`."
     default: false
   ha_proxy.stats_bind:
-    description: "Define one or several listening addresses and/or ports in a frontend."
+    description: "Define listening address and port for the stats frontend. If multithreading is enabled (`ha_proxy.threads > 1`) multiple stat pages are available - one for each thread. You can see the stat page for each thread on a separate port - starting at the defined port number."
     default: "*:9000"
   ha_proxy.stats_user:
     description: "User name to authenticate haproxy stats"
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -28,6 +28,9 @@ consumes:
     optional: true
 
 properties:
+  ha_proxy.threads:
+    description: "Optional number of threads per VM"
+    default: 1
   ha_proxy.syslog_server:
     description: "An IPv4 address optionally followed by a colon and a UDP port. It can also be an IPv6 address or filesystem path to a UNIX domain socket."
     default: "127.0.0.1"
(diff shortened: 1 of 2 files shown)


svrc-pivotal:
Great PR, but should we also include a cpu-map clause?  maybe something like
```
  <% if p("ha_proxy.threads") > 1 %>
     nbproc <%= p("ha_proxy.threads") %>
    <% for i in 1..threads do %>
     cpu-map <%= i %> <%= i-1 %>
     <% end %>
  <% end %>

```

geofffranks:
Thoughts on pulling the cpu count out of `/proc/cpuinfo`, and auto-tuning accordingly?

jgf:
@svrc-pivotal Thanks :) I tried that and measured no significant performance change. This may make more sense when used in combination with cpu pinning. When you want to pin a set amount of cpus to a certain frontend for example. In addition you need to think about what happens when you chose to have more threads than you have cpus (or less). Its definitely something that we should investigate going forward.

@geofffranks Actually one of my first tries was to put that kind of magic in the erb. However in our case the machine evaluating the erb has less cores as the one we deploy on (2 vCPUs vs 4). So at least my ad-hoc approach did not work out. Auto-tuning in general sounds nice though.

geofffranks:
Ah, that would require an auto-reconfiguration in the ctl script (bosh would render the ERB into a config template, then the ctl script would re-render that into an actual config at startup).

jgf:
So we discussed the auto-tune feature and the cpu-map suggestion.

We like the idea of having an option to use cpu-map, but we see it as a separate feature that should be enabled in combination with an option to bind threads to specific frontends - as described in http://blog.onefellow.com/post/82478335338/haproxy-mapping-process-to-cpu-core-for-maximum

Auto-tuning is a great idea, but we also need to be able to set the number of threads explicitly. For example when we only want to use 3 of 4 cores for haproxy, as we expect some load from a separate job. How about making auto-tuning optional?

For example set a specific amount of threads via
`threads: 23`
and enable auto-tuning with
`threads: "auto"`

The downside of auto-tuning is increased complexity with the template being rendered on each startup.

Whats your take on this?

geofffranks:
What about something like this in bin/haproxy_ctl to keep complexity down? We'd probably want to handle the case where no `haproxy.threads` property was provided, or default to either `1`, or `auto` in the spec file?

```
NUM_THREADS=<%= p("ha_proxy.threads") == "auto" ? "$(nproc --all)" : p("ha_proxy.threads") %>
THREAD_CONFIG=${JOB_DIR}/config/threads.conf
cat <EOF > ${THREAD_CONFIG}
global
  nbproc ${NUM_THREADS}
EOF

.......

"${DAEMON}" -f "${CONFIG}" -D -p "${PID_FILE}" -f ${THREAD_CONFIG} 0<&-
```

svrc-pivotal:
I find that more complex then just having a threads.conf.erb template that would handle nbproc and a future / separate cpu-map feature...

jgf:
Thanks, I'll try to put something together till Tuesday.

jgf:
While investigating different options for auto-tuning I stumbled across an issue regarding the stats page and socket. HAProxy keeps stats separated for each process, so in order to access the stats of all running processes we require a socket and/or a stats page per thread.

This makes things more complicated - especially when used together with auto-tuning. So I just added the changes to stats socket and pages that work for the non-auto-tuning case.

In case I want to go forward and move this functionality from the erb template to the ctl skript it can potentially become quite ugly. So should we still try to go forward with auto-tuning or do you have any suggestions regarding a simpler solution?

geofffranks:
Seems like we should skip auto-tuning for now. I'll put in an issue with a proposal for it, and if enough people want it, I'll look into implementing it.

For now, can you update the spec file's description of the stats port to reflect the new numbering if using multiple threads?

jgf:
@geofffranks Thanks, good find! I changed the documentation accordingly.

jgf:
@geofffranks Sounds great, thanks! But wait, there's more... ;)

PR #66, opened 2017-07-11
Add option to specify buffer size

Soha-Albaghdady:
This pull request adds the option to specify the buffer size to use for requests. The option enables increasing the buffer size to allow applications that require large cookies or query strings to function.


Soha-Albaghdady:
This replaces pull request https://github.com/cloudfoundry-incubator/haproxy-boshrelease/pull/64

PR #67, opened 2017-07-11
Add option to enable http health check on port 8080

Soha-Albaghdady:
The pull request adds the option to enable http health check on `haproxy_ip:8080/health`. The health check monitors the status of the http-routers. A failure is reported when all http-routers are down.


geofffranks:
Is there a need to have a health check endpoint enabled for tcp backends as well?

jgf:
That sounds like a very good idea. How about an optional health_check port setting for each tcp backend?

geofffranks:
That could work

jgf:
We are trying to add this until tomorrow.

geofffranks:
Thanks!

PR #76, opened 2017-09-16
This is a commit to let haproxy to do mutual ssl

datianshi:
The client certificate will be requested
If no cert prompt, the handshake still go through one way ssl
If a client certificate prompt, ha proxy will pass through and carry a header with cert base64 encoding
The header match what gorouter supports https://github.com/cloudfoundry-incubator/routing-release/blob/develop/jobs/gorouter/spec#L115   as x-forwarded-client-cert for mutual ssl purpose to backend apps

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -170,7 +170,18 @@ frontend https-in
     <% if p("ha_proxy.accept_proxy") %>
     bind <%= p("ha_proxy.binding_ip") %>:443 accept-proxy ssl crt /var/vcap/jobs/haproxy/config/ssl
     <% else %>
-    bind <%= p("ha_proxy.binding_ip") %>:443 ssl crt /var/vcap/jobs/haproxy/config/ssl
+    bind <%= p("ha_proxy.binding_ip") %>:443 ssl crt /var/vcap/jobs/haproxy/config/ssl <% if p("ha_proxy.client_cert") == true -%>
+      <%
+        client_ca_certs = nil
+        if_p("ha_proxy.client_ca_file") { client_ca_certs = "/var/vcap/jobs/haproxy/config/client-ca-certs.pem" }
+        unless client_ca_certs
+          client_ca_certs = "/etc/ssl/certs/ca-certificates.crt"
+        end
+      -%> ca-file <%= client_ca_certs %> verify optional <% if_p("ha_proxy.client_cert_ignore_err") do -%>
+      crt-ignore-err <%= p("ha_proxy.client_cert_ignore_err")-%> <% end -%>
+      <% if_p("ha_proxy.client_revocation_list") do -%>
+        crl-file /var/vcap/jobs/haproxy/config/client-revocation-list.pem <% end -%>
+      <% end -%>
     <% end %>
     <% if_p("ha_proxy.cidr_whitelist") do %>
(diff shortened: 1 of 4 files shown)


jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -170,7 +170,7 @@ frontend https-in
     <% if p("ha_proxy.accept_proxy") %>
     bind <%= p("ha_proxy.binding_ip") %>:443 accept-proxy ssl crt /var/vcap/jobs/haproxy/config/ssl
     <% else %>
-    bind <%= p("ha_proxy.binding_ip") %>:443 ssl crt /var/vcap/jobs/haproxy/config/ssl
+    bind <%= p("ha_proxy.binding_ip") %>:443 ssl crt /var/vcap/jobs/haproxy/config/ssl <% if p("ha_proxy.client_cert") == true %> ca-file <%= p("ha_proxy.client_cert_ca") %> verify optional crt-ignore-err all<% end %>
```
geofffranks: Can we not ignore all cert errors by default? Seems like we would want to be aware of things like a bad CA, or expired cert.

Can this support CRLs so we can check against revoked client certs as well?

datianshi:
@geofffranks Thanks for the feadback. Pushed the change you need. please review

geofffranks:
Seems like there's a missing client_ca_cert file to put the ca-file on disk as well.

 I'm also curious about the offload behavior. Right now what it looks like it does is perform mutual auth on haproxy, and then send the client cert on to the next hop for mutual auth to be done there as well. Does that need to be done in both places? Maybe the two features should be mutually exclusive, or always send the client cert on to the next hop if it was presented, and this feature just enables mutual auth inside haproxy? That way people can configure it to do auth in both places, or either of haproxy or the backend app.

datianshi:
@geofffranks They are mutual exclusively which are two different SSL termination.
The offload basically pass down an HTTP header which is layer 7 not tls. This is for go router feature (always_forward and forward). 
About the cert file:
I just referred to the default system ca certs. https://github.com/datianshi/haproxy-boshrelease/blob/backend_client_cert/jobs/haproxy/spec#L204  . In my view, it is not necessary to configure all the CA certs as params and work as a separate file. and this file can be updated through bosh trusted store.

geofffranks:
That behavior seems to differ with how mutual TLS inside CF is configured. I'd prefer to keep things consistent with their configs, and allow operators to specify the ca cert explicitly, rather than configure it one way for HAProxy, and a different way for every other CF component.

datianshi:
@geofffranks I would say. This is different with internal mutual ssl. For internal mutual ssl, basically we are using self signed cert and the NO. of Root certs are very limited and need to be configured. This is a front end certs (Not internal) It served to outside world, that should honor all the default system CA certs list (which are over hundred) trusted by public, not just a specified one. Adding a configured Self signed Root CA Cert should be prohibited for a production environment

geofffranks:
I can think of many use cases where the client-side  cert would be signed by an internal PKI system for an organization, or something that is not necessarily signed by a trusted root.

How about a `ca_pem` property, which installs the pem if specified and configures HAProxy to use that. If not specified, the release would default to using the system's root ca-file?

datianshi:
@geofffranks took your recommendation about configurable client_ca_file. If not configured it will fall back to the system certs . And rebased the branch with latest update

geofffranks:
Thanks!

PR #92, opened 2018-02-02
Fixes for keepalived stop

aveyrenc:
Checker's pid is not tracked so it's process is not stopped.

PIDs are terminated without giving them a chance to end their children so these stay running as well.


poblin-orange:
thks @aveyrenc . @geofffranks, is the next release cut planned ?

geofffranks:
Sorry, I thought I clicked the release button for v8.6.1 already. Should be out shortly

poblin-orange:
@geofffranks thks :-)

PR #94, opened 2018-03-02
allow to use crt-list configurations

dueckminor:
Hi @geofffranks, 

as @jgf already mentioned in slack,I've prepared a pull request for the crt-list feature of haproxy-1.8.x. Only things that are already supported by the haproxy-boshrelease are now also supported by crt-lists. The one and only exception is the "verify" option.
The existing "ssl_pem" feature has not been touched, but if you put both "ssl_pem" and "crt_list" in your manifest, an error will be reported.

Please have a look at my changes.    

crt-list is documented here:
  https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#5.1-crt-list

The new field 'haproxy.crt_list' replaces 'haproxy.ssl_pem' and allows to
specify sslbindconf and snifilter options for each certificate

Only a small subset of the allowed sslbindconf options is supported:

sslbindconf   | field in manifest
--------------|-------------------------
"npn"         | not supported
"alpn"        | not supported
"verify"      | "verify"
"ca-file"     | "client_ca_file"
"no-ca-names" | not supported
"crl-file"    | "client_revocation_list"
"ecdhe"       | not supported
"curves"      | not supported
"ciphers"     | "ssl_ciphers"
"ssl-min-ver" | not supported
"ssl-max-ver" | not supported

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -164,13 +164,28 @@ end
 
 <%# HTTPS Frontend %>
 
-<% if_p("ha_proxy.ssl_pem") do |pem| %>
+<%
+ssl_enabled = false
+crt_config = ""
+if_p("ha_proxy.ssl_pem") do
+    ssl_enabled = true
+    crt_config="crt /var/vcap/jobs/haproxy/config/ssl"
+end
+if_p("ha_proxy.crt_list") do
+    if ssl_enabled == true
+        abort("Please configure eihter ssl_pem or crt_list, but not both")
+    end
+    ssl_enabled=true
+    crt_config="crt-list /var/vcap/jobs/haproxy/config/ssl/crt-list"
+end
+%>
+<% if ssl_enabled == true %>
 frontend https-in
     mode http
     <% if p("ha_proxy.accept_proxy") %>
-    bind <%= p("ha_proxy.binding_ip") %>:443 accept-proxy ssl crt /var/vcap/jobs/haproxy/config/ssl <% if p("ha_proxy.strict_sni") %> strict-sni <% end %>
+    bind <%= p("ha_proxy.binding_ip") %>:443 accept-proxy ssl <%= crt_config %> <% if p("ha_proxy.strict_sni") %> strict-sni <% end %>
     <% else %>
-    bind <%= p("ha_proxy.binding_ip") %>:443 ssl crt /var/vcap/jobs/haproxy/config/ssl <% if p("ha_proxy.strict_sni") %> strict-sni <% end %> <% if p("ha_proxy.client_cert") == true -%>
+    bind <%= p("ha_proxy.binding_ip") %>:443 ssl <%= crt_config %> <% if p("ha_proxy.strict_sni") %> strict-sni <% end %> <% if p("ha_proxy.client_cert") == true -%>
       <%
         client_ca_certs = nil
         if_p("ha_proxy.client_ca_file") { client_ca_certs = "/var/vcap/jobs/haproxy/config/client-ca-certs.pem" }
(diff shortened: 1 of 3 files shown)


dueckminor:
membership publicized, feel free to close and reopen this pull request

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -164,13 +164,28 @@ end
...
+%>
+<% if ssl_enabled == true %>
 frontend https-in
     mode http
     <% if p("ha_proxy.accept_proxy") %>
-    bind <%= p("ha_proxy.binding_ip") %>:443 accept-proxy ssl crt /var/vcap/jobs/haproxy/config/ssl <% if p("ha_proxy.strict_sni") %> strict-sni <% end %>
+    bind <%= p("ha_proxy.binding_ip") %>:443 accept-proxy ssl <%= crt_config %> <% if p("ha_proxy.strict_sni") %> strict-sni <% end %>
     <% else %>
-    bind <%= p("ha_proxy.binding_ip") %>:443 ssl crt /var/vcap/jobs/haproxy/config/ssl <% if p("ha_proxy.strict_sni") %> strict-sni <% end %> <% if p("ha_proxy.client_cert") == true -%>
+    bind <%= p("ha_proxy.binding_ip") %>:443 ssl <%= crt_config %> <% if p("ha_proxy.strict_sni") %> strict-sni <% end %> <% if p("ha_proxy.client_cert") == true -%>
```
geofffranks: How do the crt-list client CA + CRLs interplay with the global property `haproxy.client_ca_file` and `ha_proxy.client_revocation_list`?

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -164,13 +164,28 @@ end
...
+<%
+ssl_enabled = false
+crt_config = ""
+if_p("ha_proxy.ssl_pem") do
+    ssl_enabled = true
+    crt_config="crt /var/vcap/jobs/haproxy/config/ssl"
+end
+if_p("ha_proxy.crt_list") do
+    if ssl_enabled == true
+        abort("Please configure eihter ssl_pem or crt_list, but not both")
```
geofffranks: Can you fix the typo on `either`?

jobs/haproxy/templates/haproxy.config.erb
dueckminor: According haproxy-1.8 documentation they override the global properties.
I've updated the `jobs/haproxy/spec` file to clarify this.

geofffranks:
Should the template compilation abort with an error if (ha_proxy.client_revocation_list || ha_proxy.client_cert || ha_proxy.ssl_pem) && haproxy.crt_list - so you can either use the old global-settings, or you can use the new SNI-specific settings, but you can't accidentally use both and run into unexpected behavior?

dueckminor:
Hmm, one the one hand you are right. But on the other hand, it's an absolutely valid configuration. You can globally define the defaults for many entries in the crt-list and override them for some entries.
I'll discuss this with my team tomorrow.

dueckminor:
I now abort the template compilation if `client_ca_file`/`client_revocation_list` is configured both globally and in a `crt-list` entry.

But it's getting more complicated...

The `ha_proxy.client_cert` option has to be configured if one of the `crt-list` entries tries to turn on mutual auth (by configuring `client_ca_file`, `client_revocation_list`, or `verify`!=`none`). Because otherwise the `X-Forwarded-Client-Cert` header will not be set.

I also found something strange in the existing `haproxy.config.erb` file which directly affects this pull request.

If we have a configuration like this:
```
properties:
  ha_proxy:
    crt_list:
    - ssl_pem: ...
      verify: required
```
mutual auth will be required and the certs from `/etc/ssl/certs/ca-certificates.crt` will be used as `ca-file`. (I don't think that configuration makes sense, but it's the simplest configuration to show the problem)

But if we now add
```
properties:
  ha_proxy:
    accept_proxy: true
```
the haproxy will not start anymore, as mutual auth is enabled but no `ca-file` is configured. As soon as `accept_proxy` is `true`, `client_ca_file` has to be configured in each crt-list entry which requires mutual authentication

Shall I fix it within this pull request?

jgf:
Regarding the `accept_proxy` and `client_cert` bug: It was introduced with this commit https://github.com/cloudfoundry-incubator/haproxy-boshrelease/commit/7778b01817df375d84a376a920004f05dbb75146#diff-dbc8ccc259666df86aec4ce94f276464L171

Where a change is missing for the `accept_proxy == true` case in line 171

geofffranks:
Let's fix the `accept_proxy` bug separately. I'll probably just add it as an optional setting that gets plugged into the bind directive instead of the current if-else syntax. Seems like the client_cert stuff needs cleaning up as well, and doesn't apply to the optional 4443 frontend.

PR #96, opened 2018-03-11
Ensure frontend options are applied appropriately to all frontends

### LINKED ISSUE #95: Mutual TLS bugs with 4443, and accept-proxy
Mutual TLS settings don't apply to the https 4443 frontend, but they should. Additionally, setting accept-proxy results in mutual TLS settings not being applied.  We should simplify the configs so it's more difficult to miss updating the TLS settings for various frontends.

geofffranks:
- The HTTP frontend now supports accept-proxy.
- Bugs where accept-proxy was not honored during mutual TLS have been
resolved
- `ha_proxy.client_cert` is no longer required to enable TLS. It is
   still honored to enable mutual tls, but the boshrelease will also
   use the presence of the following parameters to enable mutual TLS:

   - `ha_proxy.client_ca_file`
   - `ha_proxy.client_revocation_list`
   - `ha_proxy.crt_list.<i>.client_ca_file`
   - `ha_proxy.crt_list.<i>.client_revocation_list`
   - `ha_proxy.crt_list.<i>.verify` - only when value is not "none"

- The following options are now honored in the `:4443` backend:
  - `ha_proxy.cidr_whitelist`
  - `ha_proxy.cidr_blacklist`
  - `ha_proxy.block_all`
  - `ha_proxy.hsts_*`
  - `ha_proxy.rsp_headers`

- The X-Forwarded-Client_Cert now is set for requests in the `:4443`
  backend.

- The X-Forwarded-Proto header behavior in the `:4443` backend now
  matches the behavior in the `:443` backend

Fixes #95

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -49,8 +49,15 @@ if_p("ha_proxy.crt_list") do
 end
 
 client_ca_certs = "/etc/ssl/certs/ca-certificates.crt"
-if_("ha_proxy.client_ca_file") do
+if_p("ha_proxy.client_ca_file") do
   client_ca_certs = "/var/vcap/jobs/haproxy/config/client-ca-certs.pem"
+end
+
+strict_sni = ""
+if_p("ha_proxy.strict_sni") do |sni|
+  if sni == true
+    strict_sni = "strict-sni"
+  end
 end
 
 tls_options = "#{crt_config} #{strict_sni}"
@@ -59,7 +66,7 @@ if mutual_tls_enabled == true
   if_p("ha_proxy.client_cert_ignore_err") do |ignore_errs|
     tls_options = "#{tls_options} crt-ignore-err #{ignore_errs}"
   end
-  if_p("ha_proxy.client_revocation_list" do
+  if_p("ha_proxy.client_revocation_list") do
     tls_options = "#{tls_options} crl-file /var/vcap/jobs/haproxy/config/client-revocation-list.pem"
   end
 end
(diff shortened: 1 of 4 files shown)


geofffranks:
@dueckminor, @jgf can I get a sanity check on this update ? Hoping to resolve the bugs found in #94 (and a few others found in the process of resolving them), and make it less likely for updates to the TLS bind settings to be missed in certain backends

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -1,37 +1,100 @@
...
+end
+# }}}
+# TLS Bind Options {{{
+mutual_tls_enabled = p("ha_proxy.client_cert")
+ssl_enabled = false
+crt_config = ""
+if_p("ha_proxy.ssl_pem") do
+    ssl_enabled = true
+    crt_config="crt /var/vcap/jobs/haproxy/config/ssl"
+    if p("ha_proxy.client_ca_file") || p("ha_proxy.client_revocation_list")
```
dueckminor: The expected behaviour of this condition should be:
Set `mutual_tls_enabled ` to true if either `ha_proxy.client_ca_file` or `ha_proxy.client_revocation_list` is configured.

But if `ha_proxy.client_ca_file` is not configured, you get the following error:
```
      - Error filling in template 'haproxy.config.erb' (line 34: Can't find property '["ha_proxy.client_ca_file"]')
```

The following code runs without errors (but it's not really elegant)
```
    if_p("ha_proxy.client_ca_file") do
        mutual_tls_enabled = true
    end
    if_p("ha_proxy. client_revocation_list") do
        mutual_tls_enabled = true
    end
```

geofffranks:
Go for it, otherwise I can hopefully get to it in the next few days. Have been fighting fires elsewhere this week, but I appreciate you both taking a look and finding my bugs :)

jgf:
@geofffranks I'll try to come with something from our side by tomorrow

PR #135, opened 2019-04-29
Read trusted_domain_cidrs from a file

Soha-Albaghdady:
When the trusted_domain_cidrs list was too long haproxy start-up failed
with the following error:
parsing [/var/vcap/jobs/haproxy/config/haproxy.config:60]: line too long,
truncating at word 65, position 1049

Additionally trusted_domain_cidrs can be provided as a base64 encoded
string.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -20,6 +20,7 @@ templates:
   client-revocation-list.erb: config/client-revocation-list.pem
   blacklist_cidrs.txt.erb:    config/blacklist_cidrs.txt
   whitelist_cidrs.txt.erb:    config/whitelist_cidrs.txt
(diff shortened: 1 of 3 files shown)


geofffranks:
Thanks for the fix!

One question though - what was the driver for adding the base64 support?

Soha-Albaghdady:
Thanks for the review!

It only to have a shorter manifest, and also to be consistent with the `cidr_whitelist` and `cidr_blacklist`

Soha-Albaghdady:
@geofffranks thanks for merging! when is the next release planned?

geofffranks:
Should be this morning, just need to update release notes and wait on the pipelines

PR #137, opened 2019-05-08
Added errorfile template for 503

rodolfo2488:
Possible modifications:
- [X] Add configurable map for errorfiles

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,10 @@
+source 'https://rubygems.org'
+
+group :test do
+  gem 'bosh-template'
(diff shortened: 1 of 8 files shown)


geofffranks:
Thanks! How do you feel about making it a configurable error message, something like `ha_proxy.errorfiles`  as a map, with keys representing the status code to return data for, and values of the raw http responses?

rodolfo2488:
Hey @geofffranks we updated the PR with the recommended changes. Let me know if it is ok.

PR #141, opened 2019-05-23
Add cidrs_in_file feature to place cidrs in a file

gdenn:
## Type
Feature

## Description
`haproxy.cidrs_in_file` can be used to place cidrs in a file under `/var/vcap/jobs/haproxy/config/cidrs/<name>`.

Thats necessary when working with acl's that reference a long listing of cidrs. acl rules are limited in length and in production systems this limit can easily reached while working with whitelisted cidrs.

cidrs can be referenced in acls with `-f /var/vcap/jobs/haproxy/config/cidrs/<name>`, keeping the haproxy.conf cleaner and supporting large sets of cidrs.

## Major Changes
The feature is backwards compatible


gdenn:
I tried to whitelist a specific host using the this feature #138 and the `ha_proxy.http_request_deny_conditions`. The whitelist contained more then 100 cidrs which is i guess not uncommon for production systems.

I ended up receives an haproxy config error because the acl was too long. To lift the limitation i ended up placing the cidrs in a file and referencing it.

Issue #142, opened 2019-05-29
specify bpm version, sha1, url


gdenn:
Hi,

in the manifest is currently bpm with latest referenced which creates more work on the operators side because he needs to upload the latest release. Problems that occur might also be difficult to reproduce since bpm versions might differ.

I would suggest to set the bpm version specific in the manifest. We tested bpm v1.1.0 on our system.

best,
D

geofffranks:
Going to set 1.1.8 in the upcoming release as that's what's being tested against for the upcoming release.

PR #143, opened 2019-05-29
Set bpm version to v1.1.0 in manifest

gdenn:
Described in issue #142


geofffranks:
Going to set this as 1.1.8 in the upcoming release.

PR #144, opened 2019-06-25
Make HAProxy stat authentication optional

msahihi:
HAProxy stats can be enabled without authentication. 
With this PR the HAProxy-boshrelease could have the same functionality by making the user and password optional.
Additionally, authentication doesn't make sense when the service is bind to localhost and when the trusted_stats_cidrs is limited to 127.0.0.1

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -210,9 +210,9 @@ properties:
     description: "Define listening address and port for the stats frontend. If multithreading is enabled (`ha_proxy.threads > 1`) multiple stat pages are available - one for each thread. You can see the stat page for each thread on a separate port - starting at the defined port number."
     default: "*:9000"
   ha_proxy.stats_user:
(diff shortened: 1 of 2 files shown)


geofffranks:
Are you running into a problem somewhere with localhost metrics being password protected?

msahihi:
Hi @geofffranks: No, I'm not running into a serious problem. In my current implementation, I would like to configure stat without any authentication (I don't need it and the service is already bound to the localhost, not any service outside of the HAProxy) but since it is required in the release I have to define user and password.
I think it would be good to use authentication when it is required or when the credentials are set.

geofffranks:
Did more digging on this, and the default endpoint is *:9000 for stats, not localhost. This is going to negatively impact people who have `ha_proxy.stats_enable` set. However, I'm open to PRs that won't negatively harm existing deployments, or lead to easy misconfigurations and insecure posture. Feel free to re-open a new PR?

PR #148, opened 2019-07-26
add feature to avoid cross-az traffic

h0nIg:
Cross-AZ traffic will cause extra costs on iaas providers like AWS, therefore prefer traffic to local AZs.
This will be implemented via backup server, which will get called only if the remaining regular servers are down


geofffranks:
Can you make this default to being off? Otherwise, looks good

h0nIg:
@geofffranks done

PR #152, opened 2019-09-02
make bpm log path configurable

h0nIg:
since the bpm adoption, the property ha_proxy.syslog_server does not make sense in case a local unix socket from the root filesystem should get integrated. Therefore this property needs to be configurable to make the path visible inside the container

### DIFF
--- a/jobs/haproxy/templates/bpm.yml
+++ b/jobs/haproxy/templates/bpm.yml
@@ -8,9 +8,11 @@ processes:
         writable: true
       - path: /var/vcap/sys/run/haproxy
         writable: true
+<%- if p("ha_proxy.syslog_server") && p("ha_proxy.syslog_server").chars.first == "/" -%>
     unsafe:
       unrestricted_volumes:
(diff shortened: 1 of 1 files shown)


h0nIg:
pr check seem to be broken

```
Task 2 | 19:26:32 | Update stemcell: Extracting stemcell archive (00:00:12)
Task 2 | 19:26:44 | Update stemcell: Verifying stemcell manifest (00:00:00)
Task 2 | 19:26:44 | Update stemcell: Checking if this stemcell already exists (00:00:00)
Task 2 | 19:26:44 | Update stemcell: Uploading stemcell bosh-warden-boshlite-ubuntu-xenial-go_agent/456.16 to the cloud (00:01:26)
                   L Error: No space left on device @ fptr_finalize - /var/vcap/store/director/tasks/2/cpi
Task 2 | 19:28:10 | Error: No space left on device @ fptr_finalize - /var/vcap/store/director/tasks/2/cpi

```

PR #154, opened 2019-10-03
Add new feature to allow additional unrestricted volumes because BPM was blocking

Scoobed:


jobs/haproxy/templates/bpm.yml
```diff
@@ -8,11 +8,16 @@ processes:
         writable: true
       - path: /var/vcap/sys/run/haproxy
         writable: true
-<%- if p("ha_proxy.syslog_server") && p("ha_proxy.syslog_server").chars.first == "/" -%>
+
+<%- if p("ha_proxy.syslog_server") && p("ha_proxy.syslog_server").chars.first == "/"
```
geofffranks: This part is causing render failures when the syslog_server is specified, but additional_unrestricted_volumes is not.

You could make the default value `[]` instead of `~` possibly to solve this, or add a local variable for the volumes:

```
additional_volumes = []
if statement for syslog server
additional_volumes.push({"path":p("ha_proxy.syslog_server")})
...
additional_volumes.push(p("ha_proxy.additional_unrestricted_volumes"))
...
```

Scoobed: Updated the pull request as requested.

jobs/haproxy/templates/bpm.yml
```diff
@@ -8,11 +8,21 @@ processes:
...
       - path: /var/vcap/sys/run/haproxy
         writable: true
-<%- if p("ha_proxy.syslog_server") && p("ha_proxy.syslog_server").chars.first == "/" -%>
+
+<%- 
+   additional_volumes = []
+   if p("ha_proxy.syslog_server") && p("ha_proxy.syslog_server").chars.first == "/" 
+      additional_volumes.push({"path" =>p("ha_proxy.syslog_server")})
+   end 
+   if p("ha_proxy.additional_unrestricted_volumes")
```
geofffranks: because `ha_proxy.additional_unrestricted_volumes` has a default of nil, `if p("ha_proxy.additional_unrestricted_volumes")` fails because the property can't be found, so instead you would want `if_p("ha_proxy.additional_unrestricted_volumes") do`

since ha_proxy.syslog_server has a non-null default, it already exists and doesn't run into the same problem using a traditional if statement

PR #159, opened 2020-02-17
Add support for hitless reloads

domdom82:
This will enable seamless reloads of HAproxy without draining and re-deploying via BOSH.
For background info see https://www.haproxy.com/blog/hitless-reloads-with-haproxy-howto/

Background:
We would like to reload HAproxy without stopping it and losing all connections. HAproxy should only be restarted if there is a new version or a stemcell update.
Since there is no such thing as "reload" in either the BOSH lifecycle or BPM lifecycle, we will keep this as a special case and don't want to disrupt the existing deployment mechanism.

Adding this will allow us to seamlessly reload HAproxy without dropping a single connection (as long a there are no changes in the BOSH manifest of course).

I've added a signal handler to the haproxy_wrapper that reloads haproxy as described in https://www.haproxy.com/blog/hitless-reloads-with-haproxy-howto/ 

The signal can be fired conveniently by calling the added `reload` script.

The normal monit/bpm lifecycle is not affected and works just as before.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -9,6 +9,7 @@ packages:
 
 templates:
   haproxy_wrapper:              bin/haproxy_wrapper
+  reload:                       bin/reload
   drain.erb:                    bin/drain
   pre-start.erb:                bin/pre-start
   bpm.yml:                      config/bpm.yml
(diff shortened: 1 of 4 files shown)


domdom82:
@geofffranks I don't know why the CI build failed. It seems unrelated to my change.

domdom82:
@geofffranks ci build green now

geofffranks:
@domdom82  i'll try to get to this this weekend, thanks!

PR #166, opened 2020-04-06
Add support for additional external certificates

domdom82:
This PR provides the capability to add an additional source of external certificates to HAproxy.

The major use-case is to support CF-based platforms that allow their customers to register domains  and use their own certificates. These certificates may come at any time and usually happen outside the regular BOSH deploy/update lifecycle. In conjunction with the "hitless reloads" feature (see #159 ), this allows an external service to provide an additional list of certificates and dynamically load them into HAproxy without downtime.


domdom82:
@geofffranks ci failed but seems unrelated to the PR.

PR #175, opened 2021-02-11
Fix soft reload / Switch to master-worker mode

domdom82:
This PR fixes the following issues:
- `bin/reload` did no longer work for two reasons:
- Switching to BPM 1.1.9 uses [tini](https://github.com/krallin/tini) as the init process for the container and launches it without the `-g` option, which means signals are not relayed to children and instead captured by tini itself. This has stopped the `USR2` signal from propagating to the `haproxy_wrapper` script
- In [this commit](https://github.com/cloudfoundry-incubator/haproxy-boshrelease/commit/9a73a3e2c5d5c4b9386e4537eb7417bc47978b1a) we added a feature that allowed HAproxy to log to stdout, which requires launching in foreground. Unfortunately, the logic was tied to `nbproc == 1` instead of `syslog_server == stdout` which stopped the `reload` use case from working. This is because the `haproxy_wrapper` script will wait for any (blocking) child processes to finish before triggering signals like `USR2`, so the feature needed HAproxy to launch as daemon to work.

Additionally, for the `reload` use case to work with `nbproc > 1` (i.e. multi-process HAproxy) I've switched from standalone to the [master-worker mode](https://www.haproxy.com/de/blog/haproxy-process-management/) to let the master process orchestrate the reloading of each HAproxy process instead of manually killing/reloading them. This enables the `external_crt_list` feature also for deployments that use multi-processing.

During testing we have seen that reloading HAproxy while it is handling an existing connection will only move the bound sockets (e.g. `443`, `80`, `8080`, etc.) to the new instance but keep the old instance alive as long as its connections are not closed. This can lead to potentially endless instance lifetimes and over time could pile up instances handling old connections and eating all the RAM. To provide some boundaries on how many concurrent HAproxy instances can be on a system as well as how long an individual instance can live has been reloaded, two new specs have been added:
- `reload_hard_stop_after`: How long an instance can take to handle outstanding connections after reload
- `reload_max_instances`: How many concurrent instances can exist due to reloads


domdom82:
429 too many requests again. I triggered a rerun of the test.

Soha-Albaghdady (approved):

PR #187, opened 2021-05-07
New Mutual TLS setting compatible with CF Route Services

46bit:
We have Haproxy terminating TLS and then Gorouter behind Haproxy. We want to use Mutual TLS with Route Services. Right now this is broken, because Gorouter makes an external request to the route service, and Haproxy's `sanitize_set` option strips off the `X-Forwarded-Client-Cert` header.

This commit adds a new setting `forwarded_client_cert: forward_only_if_route_service`. In this mode Haproxy will preserve X-Forwarded-Client-Cert headers if there is also a X-Cf-Proxy-Signature header provided.

The security of this is based on Gorouter verifying the contents of the X-Cf-Proxy-Signature header. [Gorouter checks the contents of that header and rejects the request if it did not come from a CF Route Service](https://docs.cloudfoundry.org/services/route-services.html#x-cf-proxy-signature) (this works both for `Gorouter->Route Service` and `Route Service->Gorouter`.)

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -99,7 +99,7 @@ when "sanitize_set"
 when "forward_only_if_route_service"
   xfcc_forward_only_if_route_service = true
 else
-  abort("Unkown 'forwarded_client_cert' option: #{forwarded_client_cert_option}. Known options: 'always_forward_only', 'forward_only', 'sanitize_set'")
+  abort("Unknown 'forwarded_client_cert' option: #{forwarded_client_cert_option}. Known options: 'always_forward_only', 'forward_only', 'sanitize_set', 'forward_only_if_route_service'")
 end
 # }}}
 # IPv4 and IPv6 binding (v4v6) Option {{{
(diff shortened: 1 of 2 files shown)


jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -95,6 +96,8 @@ when "forward_only"
   xfcc_forward_only = true
 when "sanitize_set"
   xfcc_sanitize_set = true
+when "forward_only_if_route_service"
+  xfcc_forward_only_if_route_service = true
 else
   abort("Unkown 'forwarded_client_cert' option: #{forwarded_client_cert_option}. Known options: 'always_forward_only', 'forward_only', 'sanitize_set'")
```
domdom82, line 102: we should add `xfcc_forward_only_if_route_service ` to the list of possible options here

46bit:
I've followed up to get EasyCLA signed off.

This PR works to get `X-Forwarded-Client-Cert` to reach Route Services. It still doesn't get them to apps. We've filed https://github.com/cloudfoundry/routing-release/issues/203 to track that problem. I think this is still mergeable in the meantime.

plowin:
Hey @46bit , the legal parts around CLA will only be clarified during next week, we need to keep these changes open until then

46bit:
@domdom82 Me and Patrick solved the CLA issue here, so this is ready for review/merge. :)

domdom82 (approved):

PR #190, opened 2021-05-24
Implement unit tests

peterellisjones:
Add new suite of unit tests which will help avoid accidentally breaking functionality when adding future code or refactoring existing code.

Run with `bundle exec rspec` or `bundle exec guard` while developing

Screenshot 2021-05-25 at 16 08 58

Also adds linting via `bundle exec rake lint`

Some erb logic is not tested because I couldn't find a nice way to test it:
 * **haproxy_wrapper.erb** : to test this we would have execute the interpolated bash, so this is better covered by integration tests
 * **monit**: the bosh testing gem doesn't allow testing of monit files. We could test this via executing the erb directly and implementing the erb context that bosh provides, but this seemed like overkill for the small amount of logic here.
 
 I found a few potential bugs which I've marked with `FIXME` in the code. They are mostly about ensuring that users are not able to accidentally generate invalid configs.

I will raise a separate PR for integrating these unit tests and lint steps into the CI/CD pipeline


Soha-Albaghdady (approved):
LGTM 👍

PR #191, opened 2021-05-26
Run units tests in pipeline

peterellisjones:
Requires https://github.com/cloudfoundry-incubator/haproxy-boshrelease/pull/190 to be merged first

Will update pipeline to run unit tests and linting

Other changes:
* Remove notification for PR tests in slack to avoid unnecessary spam
* Limit PR tests to PRs that have been labelled with `approved`
* Fix issue with start-bosh script and move this script into the repo (was previously baked into docker image)
* Removes unused files `pipeline.yml` and `repipe`, and changes `pipeline-plain.yml` to `pipeline.yml`


ci/Dockerfile
```diff
@@ -7,4 +7,4 @@ RUN apt-get install wget
 RUN wget -q -O - https://raw.githubusercontent.com/starkandwayne/homebrew-cf/master/public.key | apt-key add -
 RUN echo "deb http://apt.starkandwayne.com stable main" | tee /etc/apt/sources.list.d/starkandwayne.list
 RUN apt-get update
-RUN apt-get install jq spruce git
+RUN apt-get -y install jq spruce git vim
```
domdom82, line 10: Is this a leftover from debugging?

ci/pipeline.yml
```diff
@@ -60,6 +60,8 @@ meta:
 groups:
   - name: (( grab meta.pipeline ))
     jobs:
+      - unit-tests
```
domdom82, line 63: I think these changes should go into `pipeline-plain.yml` which does not use spruce. We chose to not have that dependency when we took over the CI and instead go for a vanilla concourse syntax with `vars.yml` for interpolating variables.
(we should probably remove the old `pipeline.yml` as it obviously causes confusion...)

domdom82 (changes requested):

ci/Dockerfile
peterellisjones: yes but I thought we might as well leave it to make debugging easier in the future

ci/pipeline.yml
peterellisjones: I've deleted `pipeline.yml` and `repipe` and moved `pipeline-plain.yml` to `pipeline.yml`

ci/Dockerfile
domdom82: ok then. I think it won't add to much to the image.

domdom82 (approved):

PR #199, opened 2021-06-08
Add option for HTTP/2 ingress and egress

Gerg:
- Property is `ha_proxy.enable_http2`
- Defaults to disabled
- If enabled, HAProxy will use ALPN to negotiate HTTP protocol with both
clients and backends. It will prefer HTTP/2, but can negotiate down to
HTTP/1.1 for clients and backends that do not support HTTP/2.
- Only works over SSL (since ALPN is an SSL extension)

This is part of the larger effort to add HTTP/2 routing support to Cloud Foundry. See cloudfoundry/routing-release#200 for more information.

cc @belinda-liu


domdom82:
@Gerg Thanks for the PR. We are setting up acceptance tests right now. We will add a test for HTTP/2 later so you're good 😊

domdom82 (approved):

PR #206, opened 2021-06-11
HTTP2 ALPN acceptance test

peterellisjones:
~**Requires #205 is merged first**~

Adds an acceptance test for HTTP2 via https

### DIFF
--- a/acceptance-tests/https_frontend_test.go
+++ b/acceptance-tests/https_frontend_test.go
@@ -11,8 +11,18 @@ import (
 )
 
 var _ = Describe("HTTPS Frontend", func() {
-	It("Correctly proxies HTTPS requests", func() {
-		opsfileSSLCertificate := `---
+	var haproxyInfo haproxyInfo
+	var closeTunnel func()
+	var closeLocalServer func()
+	enableHTTP2 := false
+	var http1Client *http.Client
+
+	haproxyBackendPort := 12000
+	opsfileHTTPS := `---
+# Configure HTTP2
+- type: replace
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/enable_http2?
+  value: ((enable_http2))
 # Configure CA and cert chain
 - type: replace
   path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/crt_list?/-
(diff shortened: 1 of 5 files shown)


acceptance-tests/https_frontend_test.go
```diff
@@ -41,40 +52,87 @@ var _ = Describe("HTTPS Frontend", func() {
...
+		resp, err := http1Client.Get("https://haproxy.internal:443")
 		Expect(err).NotTo(HaveOccurred())
+
+		Expect(resp.ProtoMajor).To(Equal(1))
+
 		Expect(resp.StatusCode).To(Equal(http.StatusOK))
 		Eventually(gbytes.BufferReader(resp.Body)).Should(gbytes.Say("Hello cloud foundry"))
 	})
+
+	Context("When ha_proxy.enable_http2 is true", func() {
```
domdom82, line 107: There is one test missing:
"When ha_proxy.enable_http2 is false"
"It responds with a 400 bad request response to a H2 request"

domdom82 (changes requested):

acceptance-tests/https_frontend_test.go
peterellisjones: Hi @domdom82 it won't be a 400 because this happens at a lower level, but I'll add a test that verifies that an HTTP2-compatible client negotiates an HTTP 1.1 protocol when `ha_proxy.enable_http2` is false

domdom82: interesting. I thought of that initially but then I got a 400 when using openssl to test this. I noticed that curl always offers http/1.1 even if you put `--http2` so it sent a http/1.1 request when HAproxy only offered that. I think the go h2 client will do the same. 
It may be an academic point, but if you "forced" a H2 request to HAproxy that only supports HTTP/1.1 I am pretty sure it would give a 400. (H2 being binary and not at all looking like a HTTP/1.1 at all to HAproxy)

peterellisjones: Hi @domdom82 I wasn't able to find a way to cause an error with an HTTP2-compatible client -- ie seems like HAProxy happily uses HTTP2 if the client only supports HTTP2, even if `enable_http2: false`. However this test at least verifies that clients that support HTTP2 will prefer HTTP 1.1 when `enable_http2: false` https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/23a7546a494624d17fa12228b1cd188a510de4b8/acceptance-tests/https_frontend_test.go#L117

domdom82 (approved):

acceptance-tests/https_frontend_test.go
peterellisjones: Actually looks like this test is failing -- so it's using HTTP2 even without `alpn h2,http/1.1`

PR #208, opened 2021-06-14
Miscellaneous UX improvements

peterellisjones:
~**Requires #207 is merged first**~

These are a series of small changes aimed at:
1. Improving specs where there may be some confusion
2. Preventing the user from accidentally configuring properties in an inconsistent way

Changes:
* Document fact that `trusted_domain_cidrs` can be b64-encoded gzipped cidr list file
* Fix bug where TCP check port is invalid if `port` is provided but `backend_port` is not
* Removed unnecessary `then` keywords in erb templates (`then` is only needed for one-liners)
* If SSL host verification is enabled, it will error if SSL verification is not enabled
* If HSTS is disabled, it will error if HSTS preload or HSTS subdomains are set
* If `drain_frontend_grace_time` is non-zero it will error if `drain_enable` is not true
* Update `tcp` spec to document previously hidden property `backend_servers_local`
* Update `backend_prefer_local_az` spec to clarify that it only affects backend servers are provided via `http_backend` links (NOTE: this is the existing behaviour but we might want to change this in the future)
* If `client_cert` is not provided but `client_revocation_list` or `client_cert_ignore_err` are, then it will error
* If `enable_4443` (HTTPs websockets) is true, then a valid SSL config must be provided
* Update spec to clarify which types of backend `backend_crt` will be used with

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -2,36 +2,6 @@
 <%= p("ha_proxy.raw_config") %>
 <%- else -%>
 <%-
-  # Error checking
-  if !p("ha_proxy.drain_enable", false) && p("ha_proxy.drain_frontend_grace_time") > 0
-    abort "Conflicting configuration: drain_enable must be true to use drain_frontend_grace_time"
-  end
-
-  if !p("ha_proxy.hsts_enable", false)
-    if p("ha_proxy.hsts_include_subdomains")
-      abort "Conflicting configuration: hsts_enable must be true to use hsts_include_subdomains"
-    end
-    if p("ha_proxy.hsts_preload")
-      abort "Conflicting configuration: hsts_enable must be true to enable hsts_preload"
-    end
-  end
-
-  if p("ha_proxy.backend_ssl", "").downcase != "verify"
-    if p("ha_proxy.backend_ssl_verifyhost", false)
-      abort "Conflicting configuration: backend_ssl must be 'verify' to use backend_ssl_verifyhost"
-    end
-  end
-
-  if !p("ha_proxy.client_cert", nil)
-    if p("ha_proxy.client_cert_ignore_err", false)
-      abort "Conflicting configuration: must provide client_cert to use client_cert_ignore_err"
-    end
-
-    if p("ha_proxy.client_revocation_list", false)
-      abort "Conflicting configuration: must provide client_cert to use client_revocation_list"
-    end
-  end
-
 require "digest"
 # Ruby Variables to make the template more readable
 
@@ -148,6 +118,40 @@ if p("ha_proxy.enable_http2")
   alpn_config = "alpn h2,http/1.1 "
 end
 # }}}
+
+# Error checking
+  if !p("ha_proxy.drain_enable", false) && p("ha_proxy.drain_frontend_grace_time") > 0
+    abort "Conflicting configuration: drain_enable must be true to use drain_frontend_grace_time"
+  end
+
+  if !p("ha_proxy.hsts_enable", false)
+    if p("ha_proxy.hsts_include_subdomains")
+      abort "Conflicting configuration: hsts_enable must be true to use hsts_include_subdomains"
+    end
+    if p("ha_proxy.hsts_preload")
+      abort "Conflicting configuration: hsts_enable must be true to enable hsts_preload"
+    end
+  end
+
+  if p("ha_proxy.backend_ssl", "").downcase != "verify"
+    if p("ha_proxy.backend_ssl_verifyhost", false)
+      abort "Conflicting configuration: backend_ssl must be 'verify' to use backend_ssl_verifyhost"
+    end
+  end
+
+  if !p("ha_proxy.client_cert", nil)
+    if p("ha_proxy.client_cert_ignore_err", false)
+      abort "Conflicting configuration: must enable client_cert to use client_cert_ignore_err"
+    end
+
+    if p("ha_proxy.client_revocation_list", false)
+      abort "Conflicting configuration: must enable client_cert to use client_revocation_list"
+    end
+  end
+
+  if p("ha_proxy.enable_4443") && !ssl_enabled
+    abort "Conflicting configuration: if enable_4443 is true, you must provide a valid SSL config via ssl_pem or crt_list"
+  end
 -%>
 
 global
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -330,7 +330,16 @@ frontend https-in
   <%- end -%>
   <%- if p("ha_proxy.hsts_enable") -%>
     http-response set-header Strict-Transport-Security max-age=<%= p("ha_proxy.hsts_max_age").to_i %>;<% if p("ha_proxy.hsts_include_subdomains") %>\ includeSubDomains;<% end %><% if p("ha_proxy.hsts_preload") %>\ preload;<% end %>
-  <%- end -%>
+  <%-
+    else
+      if p("ha_proxy.hsts_include_subdomains")
+        abort "Conflicting configuration: hsts_enable must be true to use hsts_include_subdomains"
+      end
+      if p("ha_proxy.hsts_preload")
+        abort "Conflicting configuration: hsts_enable must be true to enable hsts_preload"
+      end
+    end
+  -%>
     capture request header Host len 256
     default_backend http-routers
   <%- if_p("ha_proxy.http_request_deny_conditions") do |conditions| -%>
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -488,6 +488,13 @@ backend http-routers
     backend_crt = "crt /var/vcap/jobs/haproxy/config/backend-crt.pem "
   end
   backend_ssl = ""
+
+  if (p("ha_proxy.backend_ssl") || "").downcase != "verify"
+    if_p("ha_proxy.backend_ssl_verifyhost") do |_|
+      abort "Conflicting configuration: backend_ssl must be 'verify' to use backend_ssl_verifyhost"
+    end
+  end
+
   if p("ha_proxy.backend_ssl").downcase == "verify"
     backend_ssl = "ssl verify required ca-file /var/vcap/jobs/haproxy/config/backend-ca-certs.pem "
     if_p("ha_proxy.backend_ssl_verifyhost") do | verify_hostname |
@@ -527,6 +534,13 @@ backend http-routed-backend-<%= prefix_hash %>
     resolvers = "resolvers default "
   end
   backend_ssl = ""
+
+  if (data["backend_ssl"] || "").downcase != "verify"
+    if data["backend_verifyhost"]
+      abort "Conflicting configuration: backend_ssl must be 'verify' to use backend_verifyhost in routed_backend_servers"
+    end
+  end
+
   if data["backend_ssl"]
     if data["backend_ssl"].downcase == "verify"
       backend_ssl = "ssl verify required ca-file /var/vcap/jobs/haproxy/config/backend-ca-certs.pem "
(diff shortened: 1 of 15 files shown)


jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -527,20 +563,27 @@ backend http-routed-backend-<%= prefix_hash %>
     resolvers = "resolvers default "
   end
   backend_ssl = ""
-  if data["backend_ssl"] then
-    if data["backend_ssl"].downcase == "verify" then
+
+  if (data["backend_ssl"] || "").downcase != "verify"
```
domdom82, line 567: why not fold this into one with the 
```
if data["backend_ssl"].downcase == "verify"
```
case in line 574?

i.e
```
if data["backend_ssl"]
  if data["backend_ssl"].downcase != "verify"
    (...)
  else
    (...)
```

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -622,22 +665,29 @@ backend tcp-<%= tcp_proxy["name"] %>
     resolvers = "resolvers default "
   end
   backend_ssl = ""
-  if tcp_proxy["backend_ssl"] then
-    if tcp_proxy["backend_ssl"].downcase == "verify" then
+
+  if (tcp_proxy["backend_ssl"] || "").downcase != "verify"
```
domdom82, line 669: Though it is shorter, I find this actually harder to read than simply
```
if tcp_proxy["backend_ssl"] && tcp_proxy["backend_ssl"].downcase != "verify"
```

I favour clarity over compactness

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -622,22 +665,29 @@ backend tcp-<%= tcp_proxy["name"] %>
...
+    if tcp_proxy["backend_verifyhost"]
+      abort "Conflicting configuration: backend_ssl must be 'verify' to use backend_verifyhost in tcp backend configuration"
+    end
+  end
+
+  if tcp_proxy["backend_ssl"]
```
domdom82, line 675: same here, we could wrap this into one `if` with line 669

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
```diff
@@ -88,23 +98,43 @@
...
+
+      context 'when client_cert is not provided' do
+        let(:properties) do
+          default_properties.merge({ 'client_cert_ignore_err' => true })
+        end
+
+        it 'aborts with a meaningful error message' do
+          expect do
+            frontend_https
+          end.to raise_error /Conflicting configuration: must provide client_cert to use client_cert_ignore_err/
```
domdom82, line 113: this should probably say: `... must enable client_cert to use ...` since `client_cert` is a boolean flag.

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
```diff
@@ -88,23 +98,43 @@
         default_properties.merge({ 'client_cert' => 'client_cert contents', 'client_cert_ignore_err' => true })
       end
 
-      # FIXME: if client_cert_ignore_err is true but client_cert is not provided, then it should error
-
       it 'adds the crt-ignore-err flag' do
         expect(frontend_https).to include('bind :443  ssl crt /var/vcap/jobs/haproxy/config/ssl  ca-file /etc/ssl/certs/ca-certificates.crt verify optional crt-ignore-err true')
       end
+
+      context 'when client_cert is not provided' do
```
domdom82, line 105: this should probably say: `when client_cert is not enabled` since client_cert is a boolean flag.

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
```diff
@@ -88,23 +98,43 @@
...
+          expect do
+            frontend_https
+          end.to raise_error /Conflicting configuration: must provide client_cert to use client_cert_ignore_err/
+        end
+      end
     end
 
     context 'when ha_proxy.client_revocation_list is provided' do
       let(:properties) do
         default_properties.merge({ 'client_cert' => 'client_cert contents', 'client_revocation_list' => 'client_revocation_list contents' })
```
domdom82, line 120: `client_cert` is actually a boolean flag. so the value should be `true` or `false` rather than a string.

See https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/master/jobs/haproxy/spec#L317

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
```diff
@@ -88,23 +98,43 @@
...
       end
 
-      # FIXME: if client_revocation_list is provided but client_cert is not provided, then it should error
-
       it 'references the crl list' do
         expect(frontend_https).to include('bind :443  ssl crt /var/vcap/jobs/haproxy/config/ssl  ca-file /etc/ssl/certs/ca-certificates.crt verify optional crl-file /var/vcap/jobs/haproxy/config/client-revocation-list.pem')
       end
+
+      context 'when client_cert is not provided' do
```
domdom82, line 127: ...not enabled

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
```diff
@@ -88,23 +98,43 @@
...
+        let(:properties) do
+          default_properties.merge({ 'client_revocation_list' => 'client_revocation_list contents' })
+        end
+
+        it 'aborts with a meaningful error message' do
+          expect do
+            frontend_https
+          end.to raise_error /Conflicting configuration: must provide client_cert to use client_revocation_list/
```
domdom82, line 135: ...must enable

spec/haproxy/templates/haproxy_config/frontend_wss_spec.rb
```diff
@@ -93,23 +100,43 @@
         default_properties.merge({ 'client_cert' => 'client_cert contents', 'client_cert_ignore_err' => true })
       end
 
-      # FIXME: if client_cert_ignore_err is true but client_cert is not provided, then it should error
-
       it 'adds the crt-ignore-err flag' do
         expect(frontend_wss).to include('bind :4443  ssl crt /var/vcap/jobs/haproxy/config/ssl  ca-file /etc/ssl/certs/ca-certificates.crt verify optional crt-ignore-err true')
       end
+
+      context 'when client_cert is not provided' do
```
domdom82, line 107: ...not enabled

spec/haproxy/templates/haproxy_config/frontend_wss_spec.rb
```diff
@@ -93,23 +100,43 @@
...
+        let(:properties) do
+          default_properties.merge({ 'client_cert_ignore_err' => true })
+        end
+
+        it 'aborts with a meaningful error message' do
+          expect do
+            frontend_wss
+          end.to raise_error /Conflicting configuration: must provide client_cert to use client_cert_ignore_err/
```
domdom82, line 115: ...must enable client_cert

spec/haproxy/templates/haproxy_config/frontend_wss_spec.rb
```diff
@@ -93,23 +100,43 @@
...
         default_properties.merge({ 'client_cert' => 'client_cert contents', 'client_revocation_list' => 'client_revocation_list contents' })
       end
 
-      # FIXME: if client_revocation_list is provided but client_cert is not provided, then it should error
-
       it 'references the crl list' do
         expect(frontend_wss).to include('bind :4443  ssl crt /var/vcap/jobs/haproxy/config/ssl  ca-file /etc/ssl/certs/ca-certificates.crt verify optional crl-file /var/vcap/jobs/haproxy/config/client-revocation-list.pem')
       end
+
+      context 'when client_cert is not provided' do
```
domdom82, line 129: ...not enabled

spec/haproxy/templates/haproxy_config/frontend_wss_spec.rb
```diff
@@ -93,23 +100,43 @@
...
+        let(:properties) do
+          default_properties.merge({ 'client_revocation_list' => 'client_revocation_list contents' })
+        end
+
+        it 'aborts with a meaningful error message' do
+          expect do
+            frontend_wss
+          end.to raise_error /Conflicting configuration: must provide client_cert to use client_revocation_list/
```
domdom82, line 137: ...must enable

domdom82 (changes requested):

jobs/haproxy/templates/haproxy.config.erb
peterellisjones: because there's an `elsif` on line 533 which selects out 'noverify' which means there is no existing case for _just_ `!= "verify"`

peterellisjones: I guess you could have 

```ruby
  if data["backend_ssl"]
    if data["backend_ssl"].downcase == "verify"
      backend_ssl = "ssl verify required ca-file /var/vcap/jobs/haproxy/config/backend-ca-certs.pem "
      if data["backend_verifyhost"]
        backend_ssl += "verifyhost #{data["backend_verifyhost"]} "
      end
      backend_ssl += alpn_config
    else 
      if data["backend_verifyhost"]
        abort "Conflicting configuration: backend_ssl must be 'verify' to use backend_verifyhost in routed_backend_servers"
      end
      if data["backend_ssl"].downcase == "noverify"
        backend_ssl = "ssl verify none "
        backend_ssl += alpn_config
      end
    end
  end
  ```
  
if you prefer.. but that's a lot of nesting.

File: jobs/haproxy/templates/haproxy.config.erb (hunk 2 above)
Comment (peterellisjones) [reply jobs/haproxy/templates/haproxy.config.erb:669]: I think it would actually need to be

```
!tcp_proxy["backend_ssl"] || (tcp_proxy["backend_ssl"] && tcp_proxy["backend_ssl"].downcase != "verify")
```

Since we still need to abort for the case where `backend_ssl` is not provided

An alternative could be `if tcp_proxy.fetch("backend_ssl", "").downcase != "verify"` if you prefer?

jobs/haproxy/templates/haproxy.config.erb
peterellisjones: This also suffers from the `elsif` on 681 meaning we'd get triple-nested ifs

The current approach does have some redundancy but it lowers the number of nested conditionals (which this file seems to have a lot of) and keeps the error checking separate

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
peterellisjones: Fixed!

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
peterellisjones: Fixed

spec/haproxy/templates/haproxy_config/frontend_wss_spec.rb
peterellisjones: Fixed

spec/haproxy/templates/haproxy_config/frontend_wss_spec.rb
peterellisjones: Fixed

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
peterellisjones: Fixed

spec/haproxy/templates/haproxy_config/frontend_wss_spec.rb
peterellisjones: Fixed

spec/haproxy/templates/haproxy_config/frontend_wss_spec.rb
peterellisjones: Fixed

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
peterellisjones: Fixed

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
peterellisjones: Fixed

jobs/haproxy/templates/haproxy.config.erb
domdom82: ok then

jobs/haproxy/templates/haproxy.config.erb
domdom82: ok let's keep it at that for now. maybe we can rework this altogether later and e.g. pull out the entire error checking into a separate section later so that we get rid of all the ifs and elses here.

domdom82 (approved):

PR #210, opened 2021-06-16
Add additional mtls headers if client_cert is used

domdom82:


peterellisjones (changes requested):
Hi @domdom82 it might be worth adding some assertions to the acceptance tests as well in (here https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/master/acceptance-tests/xfcc_test.go#L183-L187 and the other tests in that file). 

Extracting the fields from the test certificate will be something like this
```golang
block, _:= pem.Decode(creds.ClientCert.Certificate)
cert, err := x509.ParseCertificate(block.Bytes)
// ...
Expect(resp.Headers.Get("X-SSL-Client-Subject-CN")).To(Equal(cert.Subject))
```

acceptance-tests/go.sum
```diff
@@ -19,6 +19,7 @@ github.com/golang/protobuf v1.5.2/go.mod h1:XVQd3VNwM+JqD3oG2Ue2ip4fOMUkwXdXDdiu
 github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/go-cmp v0.5.5 h1:Khx7svrCpmxxtHBq5j2mp/xVjsi8hQMfNLvJFAlrGgU=
```
peterellisjones, line 22: It looks like this isn't being used --  if you run `go mod tidy` it should remove it

acceptance-tests/go.sum
```diff
@@ -79,6 +80,7 @@ golang.org/x/tools v0.0.0-20201224043029-2b0845dc783e/go.mod h1:emZCQorbCU4vsT4f
 golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
+golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 h1:go1bK/D/BFZV2I8cIQd1NKEZ+0owSTG1fDTci4IqFcE=
```
peterellisjones, line 83: It looks like this isn't being used --  if you run `go mod tidy` it should remove it

acceptance-tests/xfcc_test.go
```diff
@@ -137,7 +147,8 @@ var _ = Describe("forwarded_client_cert", func() {
 		closeLocalServer, localPort, err = startLocalHTTPServer(func(w http.ResponseWriter, r *http.Request) {
 			fmt.Println("Backend server handling incoming request")
 			recordedXFCCHeader = r.Header.Get("X-Forwarded-Client-Cert")
-			w.Write([]byte("OK"))
+			recordedHeaders = r.Header
+			_,_ = w.Write([]byte("OK"))
```
peterellisjones, line 151: Some minor whitespace issues in here and elsewhere in this file -- can be fixed with `go fmt` (VSCode's `editor.formatOnSave` will also do this automatically)

acceptance-tests/xfcc_test.go
```diff
@@ -293,3 +351,32 @@ func parseXFCCHeader(header string) string {
...
+	Expect(recordedHeaders.Get("X-Ssl-Client-Verify")).To(Equal("0"))
+	Expect(recordedHeaders.Get("X-Ssl-Client-Subject-Dn")).To(ContainSubstring(cert.Subject.Country[0]))
+	Expect(recordedHeaders.Get("X-Ssl-Client-Subject-Dn")).To(ContainSubstring(cert.Subject.Organization[0]))
+	Expect(recordedHeaders.Get("X-Ssl-Client-Subject-Dn")).To(ContainSubstring(cert.Subject.CommonName))
+	Expect(recordedHeaders.Get("X-Ssl-Client-Subject-Cn")).To(Equal(cert.Subject.CommonName))
+	Expect(recordedHeaders.Get("X-Ssl-Client-Issuer-Dn")).To(ContainSubstring(cert.Issuer.Country[0]))
+	Expect(recordedHeaders.Get("X-Ssl-Client-Issuer-Dn")).To(ContainSubstring(cert.Issuer.Organization[0]))
+	Expect(recordedHeaders.Get("X-Ssl-Client-Issuer-Dn")).To(ContainSubstring(cert.Issuer.CommonName))
+	notBeforeString := fmt.Sprintf("%02d%02d%02d%02d%02d%02dZ", cert.NotBefore.Year()-2000,
+		cert.NotBefore.Month(), cert.NotBefore.Day(), cert.NotBefore.Hour(), cert.NotBefore.Minute(), cert.NotBefore.Second())
```
peterellisjones, line 377: The not-before and not-after time comparisons could be simplified using the "magic" [Go Reference Time](https://golang.org/pkg/time/#Time.Format) on January 2nd 2006, as:
```
Expect(recordedHeaders.Get("X-Ssl-Client-Notbefore")).To(Equal(cert.NotBefore.Format("060102150405Z")))
```

(you might need `cert.NotBefore.UTC()` instead)

acceptance-tests/xfcc_test.go
```diff
@@ -293,3 +351,32 @@ func parseXFCCHeader(header string) string {
 
 	return certPEM.String()
 }
+
+func parseXFCCHeaderToCert(header string) *x509.Certificate {
```
peterellisjones, line 355: `parseXFCCHeader` should probably updated to call this new method, rather than duplicate the parsing (and maybe renamed to something more specific, like `parseXFCCHeaderToPEM`)

```
func parseXFCCHeader(header string) string {
	cert = parseXFCCHeaderToCert(header)

	var certPEM bytes.Buffer
	err = pem.Encode(&certPEM, &pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
	Expect(err).NotTo(HaveOccurred())

	return certPEM.String()
}
```

acceptance-tests/xfcc_test.go
```diff
@@ -293,3 +351,32 @@ func parseXFCCHeader(header string) string {
...
+	Expect(recordedHeaders.Get("X-Ssl-Client-Issuer-Dn")).To(ContainSubstring(cert.Issuer.Country[0]))
+	Expect(recordedHeaders.Get("X-Ssl-Client-Issuer-Dn")).To(ContainSubstring(cert.Issuer.Organization[0]))
+	Expect(recordedHeaders.Get("X-Ssl-Client-Issuer-Dn")).To(ContainSubstring(cert.Issuer.CommonName))
+	notBeforeString := fmt.Sprintf("%02d%02d%02d%02d%02d%02dZ", cert.NotBefore.Year()-2000,
+		cert.NotBefore.Month(), cert.NotBefore.Day(), cert.NotBefore.Hour(), cert.NotBefore.Minute(), cert.NotBefore.Second())
+	Expect(recordedHeaders.Get("X-Ssl-Client-Notbefore")).To(Equal(notBeforeString))
+	notAfterString := fmt.Sprintf("%02d%02d%02d%02d%02d%02dZ", cert.NotAfter.Year()-2000,
+		cert.NotAfter.Month(), cert.NotAfter.Day(), cert.NotAfter.Hour(), cert.NotAfter.Minute(), cert.NotAfter.Second())
+	Expect(recordedHeaders.Get("X-Ssl-Client-Notafter")).To(Equal(notAfterString))
+}
```
peterellisjones, line 382: VSCode: `"files.insertFinalNewline": true` :)

Otherwise can cause some problems with unix tools

domdom82: I'm using Goland but there should be a similar feature.

acceptance-tests/go.sum
domdom82: interesting. I did that and it didn't remove it.

acceptance-tests/xfcc_test.go
domdom82: seems go fmt does that already. I've turned it on in Goland and it added the newline.

acceptance-tests/xfcc_test.go
domdom82: good idea.

acceptance-tests/go.sum
peterellisjones: That's odd. If I do a fresh checkout of your branch, cd into the acceptance tests dir and run `go mod tidy` it deletes these lines (using go 1.16.3)

acceptance-tests/xfcc_test.go
domdom82: done

acceptance-tests/xfcc_test.go
domdom82: done

peterellisjones (approved):

PR #211, opened 2021-06-17
Add properties to enable HAProxy master CLI

peterellisjones:
Adds two new properties:
`master_cli_enable` (default `false`)
`master_cli_bind` (default `127.0.0.1:9001`)
With documentation in `spec` and an acceptance test

Master CLI allows managing HAProxy https://cbonte.github.io/haproxy-dconv/2.4/management.html#9.4

### DIFF
--- a/acceptance-tests/master_cli_test.go
+++ b/acceptance-tests/master_cli_test.go
@@ -1,0 +1,71 @@
+package acceptance_tests
+
+import (
+	"bytes"
+	"fmt"
+	"net"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Master CLI", func() {
+	It("Works when enabled", func() {
+		opsfileMasterCLI := `---
+- type: replace
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/master_cli_enable?
+  value: true
+- type: replace
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/master_cli_bind?
+  value: '0.0.0.0:9001'
+`
+		haproxyInfo, _ := deployHAProxy(baseManifestVars{
+			haproxyBackendPort:    12000,
+			haproxyBackendServers: []string{"127.0.0.1"},
+			deploymentName:        defaultDeploymentName,
+		}, []string{opsfileMasterCLI}, map[string]interface{}{}, true)
+
+		By("The master CLI 'show proc' command works")
+		output, err := haproxyCLICommand("show proc", fmt.Sprintf("%s:9001", haproxyInfo.PublicIP))
+		Expect(err).NotTo(HaveOccurred())
+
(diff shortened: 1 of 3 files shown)


acceptance-tests/master_cli_test.go
```diff
@@ -0,0 +1,71 @@
...
+func haproxyCLICommand(command string, addr string) (string, error) {
+	conn, err := net.DialTimeout("tcp", addr, 5*time.Second)
+	if err != nil {
+		return "", err
+	}
+	defer conn.Close()
+
+	// HAProxy master CLI does not close connection after writing response,
+	// so we will close after one second
+	if err := conn.SetDeadline(time.Now().Add(time.Second)); err != nil {
```
domdom82, line 52: shouldn't the `conn.Close()` above do this?

jobs/haproxy/templates/haproxy_wrapper.erb
```diff
@@ -84,13 +81,28 @@ trap shutdown QUIT INT TERM EXIT
...
+  end
+-%>
+
 # Start up
 echo "$(date): Starting HAProxy"
-<%- if p('ha_proxy.syslog_server') == "stdout" || p('ha_proxy.syslog_server') == "stderr" -%>
-"${DAEMON}" -f "${CONFIG}" -W -p ${PID_FILE} -V -db
-<%- else -%>
-"${DAEMON}" -f "${CONFIG}" -W -D -p ${PID_FILE}
+haproxy -f /var/vcap/jobs/haproxy/config/haproxy.config -W -p ${PID_FILE} <%= flags.join(' ') %>
```
domdom82, line 103: why keep PID_FILE in a variable but hard-code CONFIG?
I'd rather keep the CONFIG variable, too

domdom82 (changes requested):

acceptance-tests/master_cli_test.go
peterellisjones: conn.Close never gets called without the deadline. If there is no deadline, then `buffer.ReadFrom(conn)` blocks forever because the HAProxy master CLI never sends EOF after responding to the command

peterellisjones: A possible cleaner alternative would be to use https://golang.org/pkg/bufio/ to read the first 5 lines then close the connection after we get 5 since we know the response happens to be 5 lines long. However we'd still need the timeout anyway, because if for some reason we don't get 5 lines we don't want to wait forever.

jobs/haproxy/templates/haproxy_wrapper.erb
peterellisjones: NP I've changed it back. My original reasoning was that a variable that is only used once is an unnecessary abstraction. PID_FILE was used in a few different places so worth extracting into a variable.

acceptance-tests/master_cli_test.go
domdom82: ok got it

domdom82 (approved):

PR #216, opened 2021-06-25
Add test to verify that buffer_size_bytes and maxrewrite can be modified to allow large HTTP requests

peterellisjones:


acceptance-tests/buffer_size_test.go
```diff
@@ -0,0 +1,64 @@
...
+			deploymentName:        defaultDeploymentName,
+		}, []string{opsfileBufferSize}, map[string]interface{}{}, true)
+
+		closeLocalServer, localPort := startDefaultTestServer()
+		defer closeLocalServer()
+
+		closeTunnel := setupTunnelFromHaproxyToTestServer(haproxyInfo, haproxyBackendPort, localPort)
+		defer closeTunnel()
+
+		By("Sending a request to HAProxy with a 64kb header")
```
domdom82, line 36: Can we add another test that expects a 400 if the request exceeds 64k bytes? it should still work at 65k but probably throw an error at 66k, right?

peterellisjones: 66 might still be OK... I'll add one for 72kb which should definitely fail. 

The exact header limit will depend on things like the path, user agent string, and host name so I think it's better to keep a wider gap between what we expect to be allowed and what we expect to be rejected.

domdom82 (approved):

PR #219, opened 2021-07-02
All XFCC headers should respect ha_proxy.forwarded_client_cert property

peterellisjones:
Changes:

* `X-SSL-Client-*` headers describing client certificate will only be added when `ha_proxy.forwarded_client_cert` also requires that the `X-Forwarded-Client-Cert` is overwritten with the client certificate info.
* `X-SSL-Client-*` headers will be stripped from incoming requests if `ha_proxy.forwarded_client_cert` is `sanitize_set`
* Update spec to clarify when `X-SSL-Client-*` headers are added / overwritten
* Refactor `ha_proxy.config` to simplify `ha_proxy.forwarded_client_cert` implementation
* Update `ha_proxy.config` websockets  frontend `X-SSL-Client-*` header behaviour to match https frontend behaviour
* Update units tests to test new functionality and to fix following mistake: Unit tests mistakenly [stated](https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/master/spec/haproxy/templates/haproxy_config/frontend_https_spec.rb#L181) that `ha_proxy.forwarded_client_cert` is _always_forward_only_ by default, but then the test implementation tested the bahaviour of _sanitize_set_, which is the actual default. 
* Update `acceptance-tests/xfcc_test.go` to test new functionality as well as extract some shared logic to simplify test implementation

### DIFF
--- a/acceptance-tests/xfcc_test.go
+++ b/acceptance-tests/xfcc_test.go
@@ -1,16 +1,16 @@
 package acceptance_tests
 
 import (
-	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/pem"
 	"fmt"
+	"net/http"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"net/http"
 )
 
 /*
@@ -67,7 +67,7 @@ var _ = Describe("forwarded_client_cert", func() {
       common_name: haproxy.internal
       alternative_names: [haproxy.internal]
 
-# Add MTLS cert
+# Add mTLS cert
 - type: replace
   path: /variables?/-
   value:
@@ -106,17 +106,21 @@ var _ = Describe("forwarded_client_cert", func() {
 	var deployVars map[string]interface{}
 	var mtlsClient *http.Client
 	var nonMTLSClient *http.Client
-	recordedXFCCHeader := "initial"
 	var recordedHeaders http.Header
-	additionalmTLSHeaders := []string{
-		"X-Ssl-Client",
-		"X-Ssl-Client-Session-Id",
-		"X-Ssl-Client-Verify",
-		"X-Ssl-Client-Subject-Dn",
-		"X-Ssl-Client-Subject-Cn",
-		"X-Ssl-Client-Issuer-Dn",
-		"X-Ssl-Client-Notbefore",
-		"X-Ssl-Client-Notafter",
+	var request *http.Request
+
+	// These headers will be forwarded, overwritten or deleted
+	// depending on the value of ha_proxy.forwarded_client_cert
+	incomingRequestHeaders := map[string]string{
+		"X-Forwarded-Client-Cert": "my-client-cert",
+		"X-SSL-Client-Subject-Dn": "My App",
+		"X-SSL-Client-Subject-Cn": "app.mycert.com",
+		"X-SSL-Client-Issuer-Dn":  "ACME inc, USA",
+		"X-SSL-Client-Issuer-Cn":  "mycert.com",
+		"X-SSL-Client-Notbefore":  "Wednesday",
+		"X-SSL-Client-Notafter":   "Thursday",
+		"X-SSL-Client-Cert":       "ABC",
+		"X-SSL-Client-Verify":     "DEF",
 	}
 
 	AfterEach(func() {
@@ -143,10 +147,8 @@ var _ = Describe("forwarded_client_cert", func() {
 
 		By("Starting a local http server to act as a backend")
 		var localPort int
-		recordedXFCCHeader = "unknown"
 		closeLocalServer, localPort, err = startLocalHTTPServer(func(w http.ResponseWriter, r *http.Request) {
 			fmt.Println("Backend server handling incoming request")
-			recordedXFCCHeader = r.Header.Get("X-Forwarded-Client-Cert")
 			recordedHeaders = r.Header
 			_, _ = w.Write([]byte("OK"))
 		})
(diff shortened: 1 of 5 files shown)


jobs/haproxy/spec
```diff
@@ -357,12 +347,29 @@ properties:
           The client certificate received by HAProxy in the mTLS handshake will not be forwarded.
```
domdom82, line 347: We also delete the `X-SSL-Cert-xxx` headers on non-mTLS connections here. We should mention that.

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
```diff
@@ -178,68 +161,184 @@
...
+        default_properties.merge({ 'forwarded_client_cert' => 'always_forward_only' })
+      end
 
-  context 'when ha_proxy.forwarded_client_cert is forward_only' do
-    let(:properties) do
-      default_properties.merge({ 'forwarded_client_cert' => 'forward_only' })
-    end
+      it 'does not delete mTLS headers' do
+        expect(frontend_https).not_to include('http-request del-header X-Forwarded-Client-Cert')
+        expect(frontend_https).not_to include('http-request del-header X-SSL-Client')
```
domdom82, line 172: could these also be cut short using
```
expect(frontend_https).not_to include(/http-request del-header X-SSL-Client/)
```
?

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
```diff
@@ -178,68 +161,184 @@
...
+          expect(frontend_https).to include('http-request del-header X-SSL-Client-NotBefore  if !route_service_request')
+          expect(frontend_https).to include('http-request del-header X-SSL-Client-NotAfter   if !route_service_request')
+        end
 
-    it 'deletes the X-Forwarded-Client-Cert header for non-route service requests' do
-      expect(frontend_https).to include('acl route_service_request hdr(X-Cf-Proxy-Signature) -m found')
-      expect(frontend_https).to include('http-request del-header X-Forwarded-Client-Cert if !route_service_request')
-      expect(frontend_https).to include('http-request set-header X-Forwarded-Client-Cert %[ssl_c_der,base64] if { ssl_c_used }')
+        it 'overwrites mTLS headers when mTLS is used' do
+          expect(frontend_https).to include('http-request set-header X-Forwarded-Client-Cert %[ssl_c_der,base64]      if { ssl_c_used }')
```
domdom82, line 331: The new solution is a lot easier to understand, although there is one downside to it: Previous solutions only used `set-header` which implies overwriting previous instances so a `del-header` was not necessary. I still like the new way better, even if it is more verbose.

domdom82 (changes requested):

acceptance-tests/xfcc_test.go
```diff
@@ -106,17 +106,21 @@ var _ = Describe("forwarded_client_cert", func() {
...
+
+	// These headers will be forwarded, overwritten or deleted
+	// depending on the value of ha_proxy.forwarded_client_cert
+	incomingRequestHeaders := map[string]string{
+		"X-Forwarded-Client-Cert": "my-client-cert",
+		"X-SSL-Client-Subject-Dn": "My App",
+		"X-SSL-Client-Subject-Cn": "app.mycert.com",
+		"X-SSL-Client-Issuer-Dn":  "ACME inc, USA",
+		"X-SSL-Client-Issuer-Cn":  "mycert.com",
+		"X-SSL-Client-Notbefore":  "Thursday",
```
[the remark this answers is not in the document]
peterellisjones: Fixed :)

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
peterellisjones: yep!

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
peterellisjones: Are you sure? There are del-headers here that happen regardless 

https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/master/jobs/haproxy/templates/haproxy.config.erb#L361

https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/master/jobs/haproxy/templates/haproxy.config.erb#L363

https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/master/jobs/haproxy/templates/haproxy.config.erb#L369

jobs/haproxy/spec
peterellisjones: Fixed

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
domdom82: I'm not saying the old logic was great :) Only that "set-header" implies "del-header" + "add-header" but it really is easier to read now. 

Docs: http://cbonte.github.io/haproxy-dconv/2.2/configuration.html#4.2-http-request%20set-header

domdom82 (approved):

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
peterellisjones: Hi @domdom82 I've tried not to change any logic for `X-Forwarded-Client-Cert` -- if I have changed it then that is a mistake :) . There were already `del-header`s in the original implementation which is why I've kept them. You're right that set-header overrides del-header, but the set-headers are conditional (eg `if { ssl_c_used }`) which is why I assume the del-headers were added originally.

PR #220, opened 2021-07-05
Do not require SNIs when disable_domain_fronting is true

peterellisjones:
`ha_proxy.disable_domain_fronting` should not require an SNI. However when an SNI is provided, it should reject mismatched Host and SNI.

* This update also fixes the SNI check so that a Host header that includes the optional port suffix (`myhost.com:443`) is checked correctly.
* This update also adds an option to `ha_proxy.disable_domain_fronting` so that domain fronting is only checked for mtls connections

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -73,9 +73,11 @@ properties:
     default: false
   ha_proxy.disable_domain_fronting:
     description: |
+      Must be one of true, false, or "mtls_only"
       If set to true, it will prevent clients from setting a host header different from the SNI value for HTTPS and WSS (secured websockets) connections. This is called domain fronting and is mostly used by CDNs.
       If domain fronting is disabled, such requests will result in a 421 Misdirected Request error.
-      Example:
+      If set to "mtls_only", the host header will only be checked against the SNI for mtls connections
+      Example
         curl -H "Host: bob.com" https://alice.com    <-- This will result in a 421 Misdirected Request
     default: false
   ha_proxy.ssl_pem:
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -108,6 +108,11 @@ else
   abort("Unknown 'forwarded_client_cert' option: #{forwarded_client_cert_option}. Known options: 'always_forward_only', 'forward_only', 'sanitize_set', 'forward_only_if_route_service'")
 end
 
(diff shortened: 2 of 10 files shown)


jobs/haproxy/spec
```diff
@@ -73,9 +73,11 @@ properties:
     default: false
   ha_proxy.disable_domain_fronting:
     description: |
+      Must be one of "true", "false", or "mtls_only"
       If set to true, it will prevent clients from setting a host header different from the SNI value for HTTPS and WSS (secured websockets) connections. This is called domain fronting and is mostly used by CDNs.
       If domain fronting is disabled, such requests will result in a 421 Misdirected Request error.
-      Example:
+      If set to "mtls_only", the host header will only be check against the SNI for mtls connections
```
domdom82, line 79: typo `...will only be checked against...`

jobs/haproxy/spec
```diff
@@ -73,9 +73,11 @@ properties:
     default: false
   ha_proxy.disable_domain_fronting:
     description: |
+      Must be one of "true", "false", or "mtls_only"
```
domdom82, line 76: Since this is no longer a boolean value, I think the default should now be "false" instead of false. Not sure if Ruby likes mixing the type.

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -108,6 +108,11 @@ else
   abort("Unknown 'forwarded_client_cert' option: #{forwarded_client_cert_option}. Known options: 'always_forward_only', 'forward_only', 'sanitize_set', 'forward_only_if_route_service'")
 end
 
+disable_domain_fronting = p("ha_proxy.disable_domain_fronting")
+if ![true, false, "mtls_only"].include?(disable_domain_fronting)
```
domdom82, line 112: hmm does this really work? can `disable_domain_fronting` be a boolean and a string at the same time?

domdom82 (changes requested):

jobs/haproxy/spec
peterellisjones: Ruby doesn't care. But I'll add "false" and "true" (strings) as undocumented options since the intent of the user is clear in these cases

jobs/haproxy/templates/haproxy.config.erb
peterellisjones: Fixed

domdom82 (approved):

PR #222, opened 2021-07-06
Add acceptance test for ha_proxy.strict_sni

peterellisjones:


acceptance-tests/acceptance_tests_suite_test.go
```diff
@@ -64,3 +68,46 @@ func setupTunnelFromLocalMachineToHAProxy(haproxyInfo haproxyInfo, localPort, ha
...
+	Expect(err).NotTo(HaveOccurred())
+	Expect(resp.StatusCode).To(Equal(http.StatusBadRequest))
+}
+
+func expect421(resp *http.Response, err error) {
+	Expect(err).NotTo(HaveOccurred())
+	Expect(resp.StatusCode).To(Equal(http.StatusMisdirectedRequest))
+}
+
+func expectTLSUnkownCetificateAuthorityErr(err error) {
```
domdom82, line 93: typo

acceptance-tests/acceptance_tests_suite_test.go
```diff
@@ -64,3 +68,46 @@ func setupTunnelFromLocalMachineToHAProxy(haproxyInfo haproxyInfo, localPort, ha
...
 	return cancelFunc
 }
+
+func expectTestServer200(resp *http.Response, err error) {
+	Expect(err).NotTo(HaveOccurred())
+	Expect(resp.StatusCode).To(Equal(http.StatusOK))
+	Eventually(gbytes.BufferReader(resp.Body)).Should(gbytes.Say("Hello cloud foundry"))
+}
+
+func expect200(resp *http.Response, err error) {
```
domdom82, line 78: why not
```
func expectStatus(resp *http.Response, status int, err error) {
...
```
?

acceptance-tests/strict_sni_test.go
```diff
@@ -0,0 +1,129 @@
...
+		err := varsStoreReader(&creds)
+		Expect(err).NotTo(HaveOccurred())
+
+		haproxyIP := haproxyInfo.PublicIP
+
+		// Requests that use an IP rather than a hostname do not send an SNI.
+		// However the IP of HAProxy is not known until deploy time.
+		// After the first deploy, BOSH won't change the IP, so we will now
+		// redeploy and update the cert to include the IP
+		haproxyInfo, varsStoreReader = deployHAProxy(baseManifestVars{
```
domdom82, line 79: I believe there is another way to disable SNI on the client by customizing the underlying transport, so we don't have to redeploy HAproxy so often. Let's try that in a follow-up PR

domdom82 (changes requested):

acceptance-tests/strict_sni_test.go
domdom82: Found it:
When making the HTTP Client, in the transport we can set a TLS Config with servername:
```
		tlsConfig := &tls.Config{
			ServerName:         "put ip here",
                 }
```
This will be used as SNI if set to a string, but not be used if set to any (random) IP address.

acceptance-tests/acceptance_tests_suite_test.go
peterellisjones: Fixed

acceptance-tests/acceptance_tests_suite_test.go
peterellisjones: Because if it only accepts two arguments you can pass them both directly from the output of another function like this `expect200(httpClient.Get(...))`

with three arguments you would need to write
```
resp, err := httpClient.Get(...)
expectStatus(resp, err, http.StatusOK)
```

because `expectStatus(httpClient.Get(...), http.StatusOK)` is not valid Go

domdom82: Ah I see. Too bad ((a,b),c) does not unfold to (a,b,c)

acceptance-tests/domain_fronting_test.go
```diff
@@ -41,56 +42,188 @@ var _ = Describe("Domain fronting", func() {
...
+		if closeLocalServer != nil {
+			defer closeLocalServer()
+		}
+
+		if closeSSHTunnel != nil {
+			defer closeSSHTunnel()
+		}
+	})
+
+	Context("When disable domain fronting is true", func() {
```
domdom82, line 163: I think with the new host header check we now need two more tests:
- When disable domain fronting is true / mtls, it sends a 400 bad request on missing host header
- When disable domain fronting is false, it sends a 400 bad request on missing host header (this a regression test that makes sure the default behavior is unchanged)

peterellisjones: will do this as a separate PR

domdom82: Resolving: As this requires a TCP-level connection without using the http client lib, it might be more complicated. Agreed to cut a release first, then add the test.

domdom82 (approved):

acceptance-tests/acceptance_tests_suite_test.go
```diff
@@ -64,3 +68,46 @@ func setupTunnelFromLocalMachineToHAProxy(haproxyInfo haproxyInfo, localPort, ha
...
+	Expect(err).NotTo(HaveOccurred())
+	Expect(resp.StatusCode).To(Equal(http.StatusBadRequest))
+}
+
+func expect421(resp *http.Response, err error) {
+	Expect(err).NotTo(HaveOccurred())
+	Expect(resp.StatusCode).To(Equal(http.StatusMisdirectedRequest))
+}
+
+func expectTLSUnknownCetificateAuthorityErr(err error) {
```
domdom82, line 93: typo

domdom82 (changes requested):

domdom82 (approved):

PR #224, opened 2021-07-08
Acceptance test: verify that requests with no header still result in a 400 even if domain fronting is blocked

peterellisjones:
Add an acceptance test to verify that requests with no header still result in a 400 even if domain fronting is blocked

### DIFF
--- a/acceptance-tests/domain_fronting_test.go
+++ b/acceptance-tests/domain_fronting_test.go
@@ -1,10 +1,14 @@
 package acceptance_tests
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"fmt"
+	"io"
 	"net/http"
+	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -93,6 +97,7 @@ var _ = Describe("Domain fronting", func() {
 	var nonMTLSClient *http.Client
 	var mtlsClientNoSNI *http.Client
 	var nonMTLSClientNoSNI *http.Client
+	var tlsConfig *tls.Config
 	haproxyBackendPort := 12000
 
 	JustBeforeEach(func() {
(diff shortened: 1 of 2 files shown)


acceptance-tests/domain_fronting_test.go
```diff
@@ -227,3 +215,33 @@ func buildRequest(address string, hostHeader string) *http.Request {
 	req.Host = hostHeader
 	return req
 }
+
+func expect400BadRequestNoHostHeader(haproxyIP string, tlsConfig *tls.Config) {
```
domdom82, line 219: I guess this is fine for now as we have only one use case. Maybe later we can turn this into something like a `tlsClient` that is usable for generic L4 requests. Similar to the httpClients we have now.

We could then pass the client with an expected response to the `expect` function and make it more generic.

acceptance-tests/domain_fronting_test.go
```diff
@@ -171,6 +153,9 @@ var _ = Describe("Domain fronting", func() {
 			expect421(nonMTLSClient.Do(req))
 			expect421(mtlsClient.Do(req))
 
+			By("Sending a request to HAProxy with a mismatched SNI and Host header it returns a 400")
```
domdom82, line 156: Shouldn't this say rather: "Sending a request to HAproxy with no Host header it returns a 400"?
SNI check happens after the 400 so it should be unrelated.

acceptance-tests/domain_fronting_test.go
```diff
@@ -201,6 +186,9 @@ var _ = Describe("Domain fronting", func() {
 			expect200(nonMTLSClient.Do(req))
 			expect421(mtlsClient.Do(req))
 
+			By("Sending a request to HAProxy with a mismatched SNI and Host header it returns a 400")
+			expect400BadRequestNoHostHeader(haproxyInfo.PublicIP, tlsConfig)
```
domdom82, line 190: Same here. We only send a request with missing host header.

acceptance-tests/domain_fronting_test.go
```diff
@@ -171,6 +153,9 @@ var _ = Describe("Domain fronting", func() {
 			expect421(nonMTLSClient.Do(req))
 			expect421(mtlsClient.Do(req))
 
+			By("Sending a request to HAProxy with a mismatched SNI and Host header it returns a 400")
+			expect400BadRequestNoHostHeader(haproxyInfo.PublicIP, tlsConfig)
+
```
domdom82, line 158: We should have one additional test:
- domain fronting ENABLED (i.e. disableDomainFronting = false )
- Sending a request to HAproxy with no Host header it returns a 400

Just to make sure that the old logic is still intact. (beware: the old logic does not have a body so the `expect400BadRequestNoHostHeader` function will probably fail.)

The old response:
```
HTTP/1.1 400 Bad Request: missing required Host header
content-type: text/plain; charset=utf-8
connection: close

```
So the message is part of the status code instead of the body.

domdom82 (changes requested):

acceptance-tests/domain_fronting_test.go
peterellisjones: Unfortunately there isn't a tls client in the standard library -- you use a the config to directly create the connection, so we'd have to create our own.

acceptance-tests/domain_fronting_test.go
peterellisjones: fixed

acceptance-tests/domain_fronting_test.go
peterellisjones: fixed

acceptance-tests/domain_fronting_test.go
peterellisjones: I'll add it but note that this will add a few minutes to the acceptance tests

acceptance-tests/domain_fronting_test.go
domdom82: yes, I meant that. We would have to write our own little tlsClient somewhere. But let's do it only when needed.

domdom82 (approved):

PR #252, opened 2021-09-13
Added flag to separate monit healthchecks

jaristiz:
Added a flat to separate monit health checks configuration from the other
health checks, thus enabling us to test the platform on the cases that we
are starting up the foundation.

This flag is added to enable compatibility after the new monit health checks were enabled on version 10.5.0

Jhonathan Aristizabal <jhonathana@vmware.com>


crhntr:
I am working on fixing this now. Please do not merge it. Thanks :)

PR #255, opened 2021-09-14
Added a flag to separate monit healthchecks from backend health checks

crhntr:
Added a flag to separate monit healthchecks configuration from the other
healthchecks, thus enabling us to test the platform on the cases that we
are starting up.

Co-authored-by: Claire Tinati <ctinati@vmware.com>
Co-authored-by: Jhonathan Aristizabal <jhonathana@vmware.com>

### DIFF
--- a/jobs/haproxy/monit
+++ b/jobs/haproxy/monit
@@ -12,6 +12,8 @@ end
 -%>
 
 <%- if p("ha_proxy.enable_health_check_http") -%>
+
+<%- if not p("ha_proxy.disable_monit_health_check_http") -%>
 check host haproxy-health address localhost
   depends on haproxy
   if failed host localhost port <%= p("ha_proxy.health_check_port") -%> protocol http
@@ -19,4 +21,6 @@ check host haproxy-health address localhost
      with timeout <%= timeout -%> seconds
   then alert
   group vcap
+<% end -%>
+
 <% end -%>
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -190,6 +190,9 @@ properties:
   ha_proxy.enable_health_check_http:
     description: "Optionally enable http health-check on `haproxy_ip:8080/health`. It shows `200 OK` if >0 backend servers are up. If used with ext_crt_list_timeout you should make sure that the deployment canary_watch_time and update_watch_time are configured to wait at least the number of seconds defined by ext_crt_list_timeout."
     default: false
+  ha_proxy.disable_monit_health_check_http:
+    description: "The HAProxy health check endpoint returns a healthy status if at least one backend server is responding. By default when enable_health_check_http: true, Bosh will consider the HAProxy VM unhealthy if the HAProxy health check returns an unhealthy status. In some cases this might not be desired, for example when deploying HAProxy before deploying the backend servers. To prevent Bosh from considering the HAProxy VM unhealthy when all backend servers are unhealthy set disable_monit_health_check_http: true. Note that this flag is ignored unless enable_health_check_http: true."
+    default: false
   ha_proxy.health_check_port:
     description: "port for http health-check"
     default: 8080


crhntr:
@peterellisjones and/or @domdom82 please review and consider merging this. Thank you!

jobs/haproxy/spec
```diff
@@ -190,6 +190,9 @@ properties:
   ha_proxy.enable_health_check_http:
     description: "Optionally enable http health-check on `haproxy_ip:8080/health`. It shows `200 OK` if >0 backend servers are up. If used with ext_crt_list_timeout you should make sure that the deployment canary_watch_time and update_watch_time are configured to wait at least the number of seconds defined by ext_crt_list_timeout."
     default: false
+  ha_proxy.enable_monit_health_check_http:
+    description: "Optionally enable http health-check on `haproxy_ip:8080/health` using monit."
```
domdom82, line 194: This is misleading. It actually checks (using monit) that the http health endpoint above works as expected. It does not enable another http health endpoint.

peterellisjones:
Hi @crhntr thanks for the PR, we had a discussion about this and had the following concerns:

Since `enable_monit_health_check_http` is set to default to false, it will change the functionality for all existing users.

In order to avoid this can you please change the property to `disable_monit_health_check_http` and default to false.  Please also update the monit script to _only_ enable the health check if `enable_health_check_http` is true _and_ `disable_monit_health_check_http` is false. This will also allow you to revert the change to the acceptance test :) 

The logic should be:
`enable_health_check_http: true, disable_monit_health_check_http: false` =>  enable health check and verify with monit
`enable_health_check_http: true, disable_monit_health_check_http: true` => enable health check and monit ignores it
`enable_health_check_http: false, disable_monit_health_check_http: false` => (default case) disables health check and monit ignores it
`enable_health_check_http: false, disable_monit_health_check_http: true` => disables health check and monit ignores it

Could you also clarify in the description the purposes of the property ie:
1. That the property is useless unless enable_health_check_http is true
2. That the use case of the property is to ignore the backend health when making an initial deployment where the frontend is deployed _before_ the backend.

Finally, could you see if it's possible to unit test this? I'm not sure if the bosh-template gem allows you to unit test monit files, but if so we should do this

cheers,

Pete

jobs/haproxy/spec
domdom82: It should probably also mention that to retain the existing behavior, both flags should have the same value.

peterellisjones:
@domdom82 sorry didn't see your comments when I wrote https://github.com/cloudfoundry-incubator/haproxy-boshrelease/pull/255#issuecomment-920773378 -- can you confirm if the approach suggested in my comment is an OK in your opinion?

jobs/haproxy/spec
jaristiz: Hi @domdom82  the issue we have is that when we are starting up the infrastructure the monit test will fail because `HAProxy` won't report anything if the gorouter is not up yet.

jaristiz:
Hi, @peterellisjones , @domdom82 ,

Thanks for your comments, I would like more clarification about the changes, should I change the name and use the behavior mentioned by @peterellisjones? or should I change any part of the validation, in any chance I would appreciate your feedback because this change is needed for us to be able to run our CI without problems.

jaristiz:
Improved documentation for property `disable_monit_health_check_http` and changed logic as suggested.

jobs/haproxy/monit
```diff
@@ -12,6 +12,8 @@ end
 -%>
 
 <%- if p("ha_proxy.enable_health_check_http") -%>
+
+<%- if not p("ha_proxy.disable_monit_health_check_http") -%>
```
maxmoehl, line 16: Here we need to check if `ha_proxy.enable_health_check_http` is true first and if `ha_proxy.disable_monit_health_check_http` is false second.

domdom82:
@peterellisjones 

> can you confirm if the approach suggested in my comment is OK in your opinion?

my sentiments exactly :-) I just didn't have the time to write as the baby just woke up when I added the comment. I fully agree.

jaristiz:
> @peterellisjones
> 
> > can you confirm if the approach suggested in my comment is OK in your opinion?
> 
> my sentiments exactly :-) I just didn't have the time to write as the baby just woke up when I added the comment. I fully agree.

Thanks for your reply, I made all the changes as @peterellisjones suggested, would appreciate your review, and in case that everything is OK the merging.

peterellisjones:
HI @jaristiz can you clean up the commits? It looks like you've accidentally introduced changes that are not relevant to the PR -- your first two commits contain changes that were already merged with master which concern acceptance tests and SSL ciphers

jaristiz:
> HI @jaristiz can you clean up the commits? It looks like you've accidentally introduced changes that are not relevant to the PR -- your first two commits contain changes that were already merged with master which concern acceptance tests and SSL ciphers (see https://github.com/cloudfoundry-incubator/haproxy-boshrelease/pull/255/files)

Cleanup made, sorry for the inconvenience

peterellisjones:
Hi @jaristiz I think you missed this https://github.com/cloudfoundry-incubator/haproxy-boshrelease/pull/255/files#diff-52b6b825935bf6192ae4f88f9e3509974131af66dec0c413dc704432608c4507R215-R216

jobs/haproxy/spec
```diff
@@ -190,6 +190,9 @@ properties:
   ha_proxy.enable_health_check_http:
     description: "Optionally enable http health-check on `haproxy_ip:8080/health`. It shows `200 OK` if >0 backend servers are up. If used with ext_crt_list_timeout you should make sure that the deployment canary_watch_time and update_watch_time are configured to wait at least the number of seconds defined by ext_crt_list_timeout."
     default: false
+  ha_proxy.disable_monit_health_check_http:
+    description: "Optionally disable http health-check on `haproxy_ip:8080/health` using monit. This flags works in combination with enable_health_check_http, if enable_health_check_http is false the value of this property would be ignored and no checks will be performed, if enable_health_check_http is true then setting this variable as true will disable the monit checks."
```
peterellisjones, line 194: In the description can you please explain the use-case for this feature?

peterellisjones (commented):
Hi @jaristiz can you confirm that you have tested this change and that it works as expected when the new variable is `true` and `false` ?

jobs/haproxy/spec
jaristiz: Added more information about the use case for the property

jaristiz:
> Hi @jaristiz can you confirm that you have tested this change and that it works as expected when the new variable is `true` and `false` ?

Yeah we made tests yesterday.

jobs/haproxy/spec
```diff
@@ -190,6 +190,9 @@ properties:
   ha_proxy.enable_health_check_http:
     description: "Optionally enable http health-check on `haproxy_ip:8080/health`. It shows `200 OK` if >0 backend servers are up. If used with ext_crt_list_timeout you should make sure that the deployment canary_watch_time and update_watch_time are configured to wait at least the number of seconds defined by ext_crt_list_timeout."
     default: false
+  ha_proxy.disable_monit_health_check_http:
+    description: "Optionally disable http health-check on `haproxy_ip:8080/health` using monit. This switch should be used to avoid testing the healthcheck before diego is up, restoring behavior as it was before the HAProxy v10.5.0. This flags works in combination with enable_health_check_http, if enable_health_check_http is false the value of this property would be ignored and no checks will be performed, if enable_health_check_http is true then setting this variable as true will disable the monit checks."
```
peterellisjones, line 194: Hi @jaristiz I think this could still be clearer and not all users use HAProxy with Cloud Foundry. Can you change it to:

The HAProxy health check endpoint has a healthy status if at least one backend server is responding. By default when `enable_health_check_http: true`, BOSH will consider the HAProxy VM unhealthy if the HAProxy health check returns an unhealthy status. In some cases this might not be desired, for example when deploying _HAProxy_ before deploying the backend servers. To prevent BOSH from considering the HAProxy VM unhealthy when all backend servers are unhealthy set `disable_monit_health_check_http: true`. Note that this flag is ignored unless `enable_health_check_http: true`.

jaristiz: Change done.

peterellisjones (approved):

peterellisjones:
Thanks @jaristiz, LGTM. As a final item, the branch needs rebasing with master for me to be able to merge it.

crhntr:
@peterellisjones I rebased the branch on Friday. What is an ETA for a new version of HAProxy? Please let me know if there is something we can do to help.

axel7born (approved):

PR #260, opened 2021-10-07
Base64-encode X-SSL-* headers that may contain non-ASCII characters

peterellisjones:
HTTP Header are only allowed to have ASCII characters (https://stackoverflow.com/a/48138818)
X509 Certificate Properties may be UTF8 strings

When `sanitize_set` and `forward_only_if_route_service` options are used for `forwarded_client_cert`, properties from the mTLS certificate used for the frontend connection such as the common name are sent to the backend as headers such as X-SSL-Client-Subject-CN. In the case where these properties contain non-ASCII characters, invalid characters may be included in the HTTP headers, which can cause errors in backend HTTP servers.

This change base64-encodes the following headers to avoid non-ASCII characters appearing:
`X-SSL-Client-Subject-CN`
`X-SSL-Client-Subject-DN`
`X-SSL-Client-Issuer-DN`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -28,7 +28,7 @@ RSpec/MultipleMemoizedHelpers:
   Enabled: false
 
 RSpec/NestedGroups:
-  Max: 4
+  Max: 5
 
 RSpec/MultipleExpectations:
   Enabled: false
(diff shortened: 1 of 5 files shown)


domdom82:
https://en.wikipedia.org/wiki/Punycode ? instead of just dropping the characters... not sure which is more standard though

PR #269, opened 2021-11-16
Expose rise and fall healthchk options

Rob-rls:
Adds new properties `backend_health_fall` and `backend_health_rise` for backend health checks.

Currently rise and fall are not set on the backend server health checks. If the parameters are not set, HAProxy defaults the values to `fall 3 rise 2`

https://cbonte.github.io/haproxy-dconv/2.4/configuration.html#5.2-fall
https://cbonte.github.io/haproxy-dconv/2.4/configuration.html#5.2-rise

This PR will expose these options for backend health checks (anywhere where `backend_use_http_health` is set to true), while setting the default values to match the HAProxy defaults above.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -330,6 +330,8 @@ properties:
           backend_use_http_health:  true    # optional, defaults to false. enables http based health checks for the backend
           backend_http_health_port: 80      # optional, defaults to the port of the backend server. sets the port for the http backend health check
           backend_http_health_uri:  /health # optional, defaults to /health. sets the URI for backend http health checks
+          backend_health_fall: 3  # optional, defaults to 3 if not set. Number of consecutive unsuccessful health checks required before the server is considered unhealthy from a healthy state.
+          backend_health_rise: 2  # optional, defaults to 2 if not set. Number of consecutive successful health checks required before the server is considered healthy from an unhealthy state.
 
   ha_proxy.headers:
     description: "Hash of custom headers you wish you have set on each request. Spaces are automatically escaped, but any other haproxy delimiters will need to be escaped manually"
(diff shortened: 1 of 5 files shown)


jobs/haproxy/spec
```diff
@@ -330,6 +330,8 @@ properties:
           backend_use_http_health:  true    # optional, defaults to false. enables http based health checks for the backend
           backend_http_health_port: 80      # optional, defaults to the port of the backend server. sets the port for the http backend health check
           backend_http_health_uri:  /health # optional, defaults to /health. sets the URI for backend http health checks
+          backend_health_fall: 3  # optional, defaults to 3 if not set. Number of consecutive unsuccessful health checks required before the server is considered unhealthy from a healthy state.
```
peterellisjones, line 333: It's probably worth mentioning that this is ignore unless backend_use_http_health is true (and same for `backend_health_rise`)

jobs/haproxy/spec
```diff
@@ -607,6 +609,12 @@ properties:
   ha_proxy.backend_http_health_port:
     description: Port to check the backend health on
     default: 8080
+  ha_proxy.backend_health_fall:
```
peterellisjones, line 612: It's probably worth mentioning that this is ignore unless backend_use_http_health is true (and same for `backend_health_rise`)

peterellisjones (approved):

jobs/haproxy/spec
Rob-rls: Update the description to include the note about the property being ignored when backend_use_http_health is false.

jobs/haproxy/spec
Rob-rls: Update the description to include the note about the property being ignored when haproxy.backend_use_http_health is false.

PR #271, opened 2021-11-25
Additional backend acls

andy-paine:
For use as a general reverse proxy it is useful to be able to route based on more than just `path_beg`. This should be a no-op change for most users but provide a way to configure additional routing rules for backends through a new `additional_acls` property.

This PR is split into 2 commits as I wasn't sure if adding `_0` to the (default) `path_beg` acl made sense or whether to just leave it without a suffix and start numbering the additional acls only. I think I prefer handling them all the same way personally. The PR can be squashed if this is preferred.


domdom82 (approved):
Makes sense to me. Personally, I would prefer using `backend_config` and `frontend_config` to build custom ACLs and backends as I can easier read `images-backend` rather than `http-routed-backend-9c1bb7`, but I agree this PR makes the feature more usable.

domdom82:
Hey @peterellisjones adding you as reviewer, too. Maybe I missed something broken :-)

peterellisjones:
LGTM. Perhaps not as part of this PR but at some point we should now consider adding an acceptance test for routed backends, now we know there is at least one user using that feature, to help avoid any regressions in the future.

peterellisjones (approved):

PR #275, opened 2021-12-21
Parallelize acceptance tests

peterellisjones:
The acceptance tests currently take around 50 minutes to run.

The first 10 minutes of this is deploying the BOSH director using BOSH lite. This can't be easily parallelized.

The next 40 minutes is the tests themselves where we deploy HAProxy to the BOSH lite director in different configurations. This PR parallelizes this process via the `ginkgo -p` [flag](https://onsi.github.io/ginkgo/#parallel-specs) which by default uses one thread per CPU when run on a machine with four or fewer CPUs, or CPUs - 1 when run on a machine with more than four CPUs. It uses seven threads on our Concourse CI server. Note that the HAProxy deployments now need unique names per thread (although deployments in the same thread can reuse the deployment name).

This change reduces the test execution time from 40 minutes to about 20 minutes, so total time is approximately 30 minutes including the initial BOSH director setup.

Note that when deploying a new deployment the bosh director will compile any releases that haven't been compiled before with the required stemcells. To avoid all threads in the parallelized tests running the same compilations at the same time, this PR deploys HAProxy once in a `SynchronizedBeforeSuite` block which runs before any tests. This ensures that the first deployment in each of the test threads can make used of already-compiled releases.

### DIFF
--- a/acceptance-tests/acceptance_tests_suite_test.go
+++ b/acceptance-tests/acceptance_tests_suite_test.go
@@ -2,6 +2,7 @@ package acceptance_tests
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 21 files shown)


ci/release_notes.md
```diff
@@ -0,0 +1,6 @@
+# Fixes
+- acceptance tests are now run in parallel
+
+# Acknowledgements
+
+Thanks @peterellisjones for the PR
```
maxmoehl, line 6: Do we need a release for this? We don't change anything in the functionality, therefore I wouldn't create one.

acceptance-tests/acceptance_tests_suite_test.go
```diff
@@ -142,3 +169,9 @@ func checkTLSErr(err error, expectString string) {
 	Expect(errors.As(tlsErr, &opErr)).To(BeTrue())
 	Expect(opErr.Err.Error()).To(ContainSubstring(expectString))
 }
+
+func println(s string) {
+	for _, line := range strings.Split(s, "\n") {
+		fmt.Printf("node %d/%d: %s\n", ginkgoConfig.GinkgoConfig.ParallelNode, ginkgoConfig.GinkgoConfig.ParallelTotal, line)
+	}
+}
```
maxmoehl, line 177: This shadows the builtin function `println(args ...Type)`. What about just `log(s string)`?

ci/release_notes.md
maxmoehl, line 6: Do we need a release for this? We don't touch any of the actual coding therefore I wouldn't create one.

ci/scripts/acceptance-tests
```diff
@@ -41,6 +41,10 @@ bosh -n upload-stemcell $stemcell_xenial_path
 echo "----- Uploading BPM"
 bosh -n upload-release $bpm_release_path
 
+echo "----- Uploading os-conf (used for tests only)"
+bosh -n upload-release --sha1 386293038ae3d00813eaa475b4acf63f8da226ef \
+  https://bosh.io/d/github.com/cloudfoundry/os-conf-release?v=22.1.2
+
```
maxmoehl, line 47: Why was this added now? I thought this doesn't change any of the tests? And would it be possible to consume the other releases from bosh.io directly as well?

maxmoehl (changes requested):
Looks good overall

ci/release_notes.md
peterellisjones: Good point

acceptance-tests/acceptance_tests_suite_test.go
peterellisjones: We can't use `log` because that is a package in the stdlib which is imported in eg acceptance-tests/remote.go

peterellisjones: The stdlib function is fmt.Println FYI so it's not shadowed unless you import `fmt` as a dot import (which we're not doing anywhere)

I've changed the function name to `print` to avoid confusion with `fmt.Println`

ci/scripts/acceptance-tests
peterellisjones: Just a small refactor so that all releases are uploaded in the same place before running the tests. This was previously being added via a URL in https://github.com/cloudfoundry/haproxy-boshrelease/blob/master/acceptance-tests/bosh_helpers.go#L50-L57 

>  And would it be possible to consume the other releases from bosh.io directly as well?
yes you could consume the BPM release directly from bosh.io however since BPM is also included in the final bosh release tarball, doing it via the pipeline makes it easier to track which version we have

ci/release_notes.md
peterellisjones: good point, removed

acceptance-tests/acceptance_tests_suite_test.go
maxmoehl: I'm talking about the builtin function, not the stdlib. You can call this without any package prefix and you can shadow them by defining something else with the same name. But IMO that's confusing and should be avoided. See https://pkg.go.dev/builtin#println (print is also a builtin).

peterellisjones: Ah never even heard of that.

I've changed it to `writeLog`

maxmoehl (approved):
LGTM

PR #276, opened 2021-12-23
Fix HAProxy regex support regression

peterellisjones:
**Regex based ACLs were not working at all**. The regex engine for HAproxy, PCRE2 was silently failing, causing all regex-based ACLs to not match. Adding the --enable-jit option when compiling PCRE2 fixes the issue.

This PR also adds an acceptance test to catch future regressions.


maxmoehl (approved):
LGTM

PR #278, opened 2022-02-15
Add SHA-256 for the Release page

peanball:
Replaces the SHA1 key for releases on the release page with the SHA256 hash.

### DIFF
--- a/ci/scripts/shipit
+++ b/ci/scripts/shipit
@@ -68,6 +68,8 @@ bosh -n create-release releases/$RELEASE_NAME/$RELEASE_NAME-$VERSION.yml \
 cd -
 
 RELEASE_TGZ=$REPO_ROOT/releases/$RELEASE_NAME/$RELEASE_NAME-$VERSION.tgz
+export SHA1=$(sha1sum $RELEASE_TGZ | head -n1 | awk '{print $1}')
+echo "SHA1=$SHA1"
 export SHA256=$(sha256sum $RELEASE_TGZ | head -n1 | awk '{print $1}')
 echo "SHA256=$SHA256"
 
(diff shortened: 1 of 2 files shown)


ci/scripts/shipit
```diff
@@ -81,10 +81,10 @@ cat >> ${RELEASE_ROOT}/notes.md <<EOF
...
 \`\`\`yaml
 releases:
-- name: $RELEASE_NAME
-  version: $VERSION
-  url: https://github.com/${GITHUB_OWNER}/${RELEASE_NAME}-boshrelease/releases/download/v${VERSION}/${RELEASE_NAME}-${VERSION}.tgz
-  sha1: $SHA1
+- name: "$RELEASE_NAME"
+  version: "$VERSION"
+  url: "https://github.com/${GITHUB_OWNER}/${RELEASE_NAME}-boshrelease/releases/download/v${VERSION}/${RELEASE_NAME}-${VERSION}.tgz"
+  sha256: "sha256:$SHA256"
```
axel7born, line 87: We shouldn't replace sha1 with sha256.
Rather, we should add sha256 since sha1 might be useful for other consumers.

axel7born (changes requested):

ci/scripts/shipit
peanball: Changing the key to `sha256` from `sha1` was a mistake. I will check where this is used and if that supports a `sha256` key as well.

If it does I will put both, if not I will put the syntax similar to what we have in the artifacts.yml file, i.e.:

```yaml
   sha1: "sha256:${SHA256}"
```

peanball: The key `sha256` is not supported. A sha256 hash can only be provided via the syntax as mentioned above, with the `sha256:` prefix.

All public bosh releases we found use `sha1`, so we add the `sha256:` notation in a commented out field for those users that may benefit from it but keep the `sha1` hash as default.

axel7born (approved):

PR #279, opened 2022-03-01
2.5.1

Rob-rls:
Update the HAProxy version to 2.5.1

The upgrade required changing how the 'grace' keyword is used within the haproxy config.  Previously it was added to each indvidual frontend, now it must be enabled as a global option.  See the blow documentation from HAProxy for details.

https://www.haproxy.com/blog/announcing-haproxy-2-5/#graceful-stop-of-all-proxies
https://cbonte.github.io/haproxy-dconv/2.5/configuration.html#grace

### DIFF
--- a/config/blobs.yml
+++ b/config/blobs.yml
@@ -1,7 +1,6 @@
-haproxy/haproxy-2.4.4.tar.gz:
-  size: 3587306
-  object_id: 5ff07959-cb68-4de2-59c1-d9886ca1d8a6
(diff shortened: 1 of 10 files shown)


peanball:
Changes look good. Only comment: HAProxy 2.5.4 was released on 2022/02/25 [see Changelog](http://www.haproxy.org/download/2.5/src/CHANGELOG). Should we consider that latest version instead?

Rob-rls:
closing this PR.  made a new one for 2.5.4

Issue #284, opened 2022-04-04
HAProxy leaves null bytes in the log file after log rotation


mariash:
When using default syslog_server value as `stdout` we are seeing that HAProxy log file `haproxy.stdout.log` keeps growing even after logs are rotated and there are null bytes at the beginning of the file.

To reproduce:

1. Generate some logs, e.g. send HTTP requests to the deployed app
2. On HAProxy VM rotate logs: `logrotate -f -v /etc/logrotate.conf`
3. See that HAProxy log is 0 KB.
4. Generate more logs
5. See that the size of HAProxy log jumped to include previous size.
6. See that the `haproxy.stdout.log` contains null bytes at the beginning of the file.

We believe that the issue is that HAProxy process keeps writing at the previous offset even though the file was truncated.

It seems like you can add a pipe when starting haproxy letting cat to handle the logging eventually instead of haproxy, e.g.

`haproxy -f ${CONFIG} -W -p ${PID_FILE} -V -db | cat –`

maxmoehl:
Thanks for bringing this to our attention! I reproduced this issue with a minimal docker-compose setup and found out that the issue is not related to haproxy. The issue is that the redirection must happen with the `O_APPEND` flag (see also [this question](https://serverfault.com/questions/510469/null-bytes-in-a-file-created-by-2-rotated-with-logrotate)), so instead of doing `haproxy > stdout.log` we must use `haproxy >> stdout.log`. This fix works in the local testing environment I built.

Next step is to figure out where the redirection happens for our release because it's not inside our script.

domdom82:
I've merged #285 as an intermediate fix to unblock this issue until we find a more elegant (correct?) solution. @maxmoehl Let's aim for a new release this week.

PR #285, opened 2022-04-04
Resolve the null bytes in log file after rotation

mariash:
When logs are rotated with logrotate using copytruncate haproxy does not
reset the offset and keeps writing from the old offset prepending logs
with null bytes. This results in HAProxy logs growing indefinetely.

Redirect output to pipe and let cat write to log file.

#284

### DIFF
--- a/jobs/haproxy/templates/haproxy_wrapper.erb
+++ b/jobs/haproxy/templates/haproxy_wrapper.erb
@@ -84,12 +84,14 @@ trap reload_daemon USR2
 
 <%-
   flags=[]
+  cat_pipe=''
 
   # Enable background mode only if syslog is not configured to output to stdout/stderr
   run_in_foreground = p('ha_proxy.syslog_server') == "stdout" || p('ha_proxy.syslog_server') == "stderr"
 
   if run_in_foreground
     flags += ['-V', '-db']
+    cat_pipe='| cat -'
   else
     flags += ['-D']
   end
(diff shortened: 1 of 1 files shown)


jobs/haproxy/templates/haproxy_wrapper.erb
```diff
@@ -101,7 +101,7 @@ trap reload_daemon USR2
 
 # Start up
 echo "$(date): Starting HAProxy"
-haproxy -f ${CONFIG} -W -p ${PID_FILE} <%= flags.join(' ') %>
+haproxy -f ${CONFIG} -W -p ${PID_FILE} <%= flags.join(' ') %> | cat -
```
domdom82, line 104: @mariash this only applies to when HAproxy is writing to stdout, right? can you wrap this pipe in a guard condition for that? 
Along the lines of line 91 where we add flags based on the `syslog_server` setting. You could introduce a `cat_pipe` variable that is `| cat -` if `run_in_foreground` and empty if not.

domdom82, line 104: @mariash this only applies to when HAproxy is writing to stdout, right? can you wrap this pipe in a guard condition for that? 
Along the lines of line 91 where we add flags based on the `syslog_server` setting. You could introduce a `cat_pipe` variable that is `| cat -` if `run_in_foreground` and empty if not.

mariash:
@domdom82 Hey, I updated the PR with your suggestion, please review.

domdom82 (approved):

PR #286, opened 2022-04-06
Wait for the main HAProxy process in daemon mode

peanball:
This waits for the main HAProxy process' PID instead of waiting indefinitely.
When the main process goes down, HAProxy will be restarted automatically by BPM.

### DIFF
--- a/jobs/haproxy/templates/haproxy_wrapper.erb
+++ b/jobs/haproxy/templates/haproxy_wrapper.erb
@@ -105,6 +105,7 @@ haproxy -f ${CONFIG} -W -p ${PID_FILE} <%= flags.join(' ') %>
 
 <%- if !run_in_foreground -%>
 # Since HAProxy requires Daemonized mode for its `nbproc` multi-process feature
-# to work properly, and BPM requires the process to stay running, sleep infinitely here
-sleep infinity
+# to work properly, and BPM requires the process to stay running, wait for the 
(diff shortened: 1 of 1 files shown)


jobs/haproxy/templates/haproxy_wrapper.erb
```diff
@@ -105,6 +105,7 @@ haproxy -f ${CONFIG} -W -p ${PID_FILE} <%= flags.join(' ') %>
 
 <%- if !run_in_foreground -%>
 # Since HAProxy requires Daemonized mode for its `nbproc` multi-process feature
-# to work properly, and BPM requires the process to stay running, sleep infinitely here
-sleep infinity
+# to work properly, and BPM requires the process to stay running, wait for the 
+# HAProxy main process to terminate
+while kill -0 $(cat ${PID_FILE}) 2>/dev/null; do sleep 1; done;
```
domdom82, line 110: so this basically says: "check every second if we can send signals to the haproxy pid." `kill` will throw an error if the pid is gone so that would break the look. right? 
I think the same could be achieved by simply calling `pgrep` or maybe even `ps`?

peanball: Yes exactly.
You could get the same with `pgrep` (if it's installed) and ps (if it's not on `busybox`). Kill always works and is there.

PR #291, opened 2022-05-11
Add upstream version numbers for contained components in release information

peanball:
The upstream release version (e.g. HAProxy) contained in an `haproxy-boshrelease` are not evident from its version number.

This change introduces a table of version numbers contained in a particular boshrelease. The version numbers are automatically taken from `packages/*/packaging` and should be representative.


ci/scripts/shipit
```diff
@@ -78,7 +78,23 @@ echo "v${VERSION}"                         > ${RELEASE_ROOT}/tag
...
+### Versions
+
+The following versions of upstream components are included in this haproxy-boshrelease:
+
+| Component | Version |
+| ----------| ------- |
+| HAProxy   | \`${HAPROXY_VERSION}\` |
+| PCRE      | \`${PCRE_VERSION}\` |
+| Lua       | \`${LUA_VERSION}\` |
+| socat     | \`${SOCAT_VERSION}\` |
```
domdom82, line 97: looks like this is missing keepalived?

peanball: There was no explicit version number indication in the keepalived component, but I will extract it into a variable and put it here.

domdom82 (approved):

PR #296, opened 2022-07-12
HAProxy 2.5.7 and `h1-accept-payload-with-any-method`

maxmoehl:
This PR bumps HAProxy to version 2.5.7 and replaces the patch to allow bodies when using HTTP/1.0 with the proper option introduced with 2.5.7.


spec/haproxy/templates/haproxy_config/global_and_default_options_spec.rb
```diff
@@ -491,4 +491,16 @@
...
+    let(:properties) do
+      {
+        'always_allow_body_http10' => true
+      }
+    end
+
+    it 'sets the global option' do
+      expect(global).to include('h1-accept-payload-with-any-method')
+    end
+  end
```
peanball, line 505: I would expect tests more akin to [the regex fix, where we use actual requests and verify the behaviour / response codes](https://github.com/cloudfoundry/haproxy-boshrelease/pull/276/files#diff-9edf578784347fdf4f72aa066f00020fe3f84d2f04d0e33aa2c1801fff961372)

peanball: right... ok.

for reference:
```go
        // The protocol version for incoming server requests.
	//
	// For client requests, these fields are ignored. The HTTP
	// client code always uses either HTTP/1.1 or HTTP/2.
	// See the docs on Transport for details.
	Proto      [string](https://pkg.go.dev/builtin#string) // "HTTP/1.0"
	ProtoMajor [int](https://pkg.go.dev/builtin#int)    // 1
	ProtoMinor [int](https://pkg.go.dev/builtin#int)    // 0
```

maxmoehl:
I'm still testing this change on one of our landscapes and will merge afterwards

maxmoehl:
Verified, works as intended. Waiting for the acceptance-tests to finish

PR #298, opened 2022-07-15
Implement Auto-Bumping of Dependencies

a18e:


ci/pipeline.yml
```diff
@@ -12,115 +11,116 @@ groups:
...
       - major
       - minor
+      - autobump-dependencies
 
 jobs:
   - name: unit-tests
     public: true
     serial: true
     plan:
-    - do:
```
domdom82, line 21: shouldn't this change only add
- autobump-job
- daily trigger
?
seems like there is a whole lot of useless whitespace formatting going on here...

domdom82 (changes requested):

domdom82 (approved):

PR #313, opened 2022-08-04
Lua acceptance test

peanball:
Add an integration test that installs and checks a small Lua service (applet) that returns the current Lua version.
The Lua version is also output to the test's result outputs as provided by the server.

### DIFF
--- a/acceptance-tests/lua_test.go
+++ b/acceptance-tests/lua_test.go
@@ -4,24 +4,24 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 )
 
 var _ = FDescribe("Lua scripting", func() {
 	It("Deploys haproxy with lua script", func() {
-		replyLuaTargetPath := "/var/vcap/packages/haproxy/lua_test.lua"
-		opsfileLua := fmt.Sprintf(`---
+		opsfileLua := `---
 # Enable Lua scripting
 - type: replace
   path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/lua_scripts?/-
-  value: "%s"
+  value: "/var/vcap/packages/haproxy/lua_test.lua"
 - type: replace
   path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/frontend_config?
   value: |-
     acl lua_path path /lua_test
-    http-request use-service lua.lua_test if lua_path
-`, replyLuaTargetPath)
+    http-request use-service lua.lua_test if lua_path`
 
 		replyLuaContent := `
 local function lua_test(applet)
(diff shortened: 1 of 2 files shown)


acceptance-tests/lua_test.go
```diff
@@ -0,0 +1,69 @@
...
+		haproxyInfo, _ := deployHAProxy(baseManifestVars{
+			haproxyBackendPort:    haproxyBackendPort,
+			haproxyBackendServers: []string{"127.0.0.1"},
+			deploymentName:        deploymentNameForTestNode(),
+		}, []string{opsfileLua}, map[string]interface{}{}, true)
+
+		// upload Lua script file
+		uploadFile(haproxyInfo, strings.NewReader(replyLuaContent), replyLuaTargetPath)
+
+		closeLocalServer, localPort := startDefaultTestServer()
```
domdom82, line 57: you might not even need a test server because you don't need a backed. the health check will be down but you can ignore that when calling lua.

domdom82 (changes requested):

acceptance-tests/lua_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"net/http"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+)
+
+var _ = Describe("Lua scripting", func() {
+	It("Deploys haproxy with lua script", func() {
+		replyLuaTargetPath := "/var/vcap/packages/haproxy/lua_test.lua"
+		opsfileLua := fmt.Sprintf(`---
```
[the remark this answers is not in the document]
peanball: it's reused in another place. I didn't want to have it twice.

acceptance-tests/lua_test.go
peanball: It won't harm either, correct? I'm not sure how well the setup handles not having the backend and it's difficult to test locally right now.

domdom82: it makes the test longer. not much but it counts.
just now this:
the way you are starting haproxy now it will report unhealthy anyways until the healthchecks to the backend had a chance to run a couple times. so without the `eventually` closure like in other tests, by the time your code (calling lua) runs, haproxy will still be unhealthy. so adding the server without waiting for health to turn green is actually pointless.

peanball: ok, fair. I was worried it would not start HAProxy if it can't reach the backend. Not starting is only the case when it can't _resolve_ the address though. I will take it out.

peanball: For completeness: the backend is back in as there is no way in the current test harness to disable it (easily). We will look into having an HAProxy without backend in the future.

b1tamara (approved):
LGFM

acceptance-tests/lua_test.go
```diff
@@ -0,0 +1,79 @@
...
+		// and monit should in turn start reporting a healthy process
+		// We will up to wait one minute for the status to stabilise
+		Eventually(func() string {
+			return boshInstances(deploymentNameForTestNode())[0].ProcessState
+		}, time.Minute, time.Second).Should(Equal("running"))
+
+		By("Sending a request to HAProxy with Lua endpoint")
+		resp, err := http.Get(fmt.Sprintf("http://%s/lua_test", haproxyInfo.PublicIP))
+		
+		expectLuaServer200(resp, err)
```
domdom82, line 75: you can actually merge this into
```suggestion
		expectLuaServer200(http.Get(fmt.Sprintf("http://%s/lua_test", haproxyInfo.PublicIP)))
```

just sayin' :)

domdom82 (approved):

acceptance-tests/lua_test.go
peanball: It looks more legible the other way around and I reference `resp` in the line below.

The output is for visual inspection of the actual Lua version. It's not easily possible to make a correct expectation to check, which string Lua will report as version, so it is an output for interested parties that look at the build log.

@b1tamara had asked me about this line but I would prefer to keep it.

This means I will not shorten the call either.

PR #318, opened 2022-09-11
Fix local docker mac acceptance

peanball:
Docker for Mac changed to using cgroups v2 by default.

These instructions help set it back to v1, as BOSH lite and the docker-cpi do not run with cgroups v2.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -405,7 +405,7 @@ jobs:
                 type: docker-image
                 source:
                   repository: iacbox.common.repositories.cloud.sap/haproxy-boshrelease-testflight
-                  tag: "2022-08-03"
+                  tag: latest
                   username: ((docker.username))
                   password: ((docker.password))
               run:
--- a/acceptance-tests/README.md
+++ b/acceptance-tests/README.md
@@ -31,6 +31,12 @@ A convenience script that does this for you is below.
 docker_restart_with_cgroupsv1() {
     SETTINGS=~/Library/Group\ Containers/group.com.docker/settings.json
 
+    if ! command -v jq >/dev/null || ! command -v sponge; then
+        echo "Requires jq and sponge. Consider installing via:"
+        echo "   brew install jq moreutils"
+        return
+    fi
+
     cgroupsV1Enabled=$(jq '.deprecatedCgroupv1' "$SETTINGS")
     if [ "$cgroupsV1Enabled" = "true" ]; then
         echo "deprecatedCgroupv1 is already set to 'true'. Acceptance tests should work."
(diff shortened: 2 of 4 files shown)


ci/Dockerfile
```diff
@@ -14,6 +14,20 @@ RUN apt-get update && apt-get install -y wget jq git vim nano python3-pip\
 COPY scripts/requirements.txt /requirements.txt
 RUN /usr/bin/python3 -m pip install -r /requirements.txt
 
+RUN echo "# Use local bosh release -- useful for testing \
```
domdom82, line 17: why not make this a local ops file and copy it it to `/usr/local/local-releases.yml` as a COPY step instead of this inline RUN step? would make the Dockerfile less clustered

ci/pipeline.yml
```diff
@@ -405,7 +405,7 @@ jobs:
                 type: docker-image
                 source:
                   repository: iacbox.common.repositories.cloud.sap/haproxy-boshrelease-testflight
-                  tag: "2022-08-03"
+                  tag: latest
```
domdom82, line 408: this could be dangerous. since we don't build the image regularly and push it to the registry, it will grow stale over time and we won't know what "latest" means when we look at it. we can't compare it to the actual latest from upstream then.
I'd rather opt for
a) keep the tag but bump it
b) keep "latest" but rebuild and push the image regularly

domdom82 (changes requested):

ci/pipeline.yml
peanball: I'm for b) and this is what I started now. `latest` is now in fact a fresh image (making sure it actually works properly at all)

My longer-term plan is to create a new concourse job that would create a new image when there are e.g. new stem cells and possibly once a week or something like that. This is ideally in combination with a 'canned director state' that could just be revived instead of building it up from scratch on every run.

ci/Dockerfile
peanball: Yes that's fair. Will do. This file at some point disappeared form the `main-bosh-docker` parent image. I added this as quick workaround to keep going and it stayed like this.

ci/pipeline.yml
domdom82: Agreed.

ci/Dockerfile
kecirlotfi: You are calling `sponge` command which is not available by default. We should install it before running the script.

peanball: > You are calling `sponge` command which is not available by default. We should install it before running the script. 

We might not even need it. I'll test with a simple redirect and add a change.

peanball: I've removed the file altogether as it wasn't really used. The file that was written in the Dockerfile was actually not consistent (i.e. it referenced a variable that is never set) but was seemingly ignored.
The newer versions of our base image `bosh/main-bosh-docker` don't have the local-releases.yml opsfile anymore. I've removed it from our copy of the `start-bosh` script as well now. We should consider reverting back to the `start-bosh` script of the underlying image again at some point.

plowin (approved):
lgtm

PR #320, opened 2022-09-11
Bump rubocop used for lining

peanball:
Update the linter to a newer version. The old version seems to have issues with the update base TestFlight image.

Also addresses new warnings from the new linter.


peanball:
merged with #319

PR #325, opened 2022-09-15
Keep only the version number for the release note, not the comment after

peanball:
With the autobumper we've added the URL origins for specific versions after the version number. The release notes appender in `shipit` did not filter those out and created a messy Release page.

This fixes the version number extraction to include only the version number, not the comment behind it.


b1tamara (approved):
LGTM

PR #331, opened 2022-10-10
Add Idle close on response option

domdom82:
This should prevent idle connections from being terminated immediately during restarts/reloads.

The default is to enable to option as this used to be the default behavior prior to 2.4 and was changed silently.

We also have a default value for hard-stop-after so enabling the option by default should be safe.

Changes:

- idle-close-on-response enabled by default (HAproxy will gracefully close idle conns now, instead of closing them outright)
- draining will no longer cause HAproxy to be restarted by monit after it has terminated itself
- the draining process is now handled completely within the drain script. BOSH wait time is set to 0.
- HAproxy will still be restarted after crashes unrelated to draining
- made the unit tests for the drain script less strict and error-prone

### DIFF
--- a/acceptance-tests/drain_test.go
+++ b/acceptance-tests/drain_test.go
@@ -2,7 +2,10 @@ package acceptance_tests
 
 import (
 	"fmt"
+	"io"
+	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -27,6 +30,10 @@ var _ = Describe("Drain Test", func() {
 - type: replace
   path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/drain_timeout?
   value: 10
+# Increase idle timeout between requests
+- type: replace
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/keepalive_timeout?
+  value: 60
 `
 	It("Honors grace and drain periods", func() {
 		haproxyBackendPort := 12000
@@ -69,14 +76,16 @@ var _ = Describe("Drain Test", func() {
 		time.Sleep(10 * time.Second)
 
 		By("After grace period has passed, draining should set in, disabling listeners")
-		_, err = http.Get(fmt.Sprintf("http://%s:8080/health", haproxyInfo.PublicIP))
+		// We need a new client so there won't be any reusable connections
+		httpClient := &http.Client{}
+		_, err = httpClient.Get(fmt.Sprintf("http://%s:8080/health", haproxyInfo.PublicIP))
 		expectConnectionRefusedErr(err)
-		_, err = http.Get(fmt.Sprintf("http://%s", haproxyInfo.PublicIP))
+		_, err = httpClient.Get(fmt.Sprintf("http://%s", haproxyInfo.PublicIP))
 		expectConnectionRefusedErr(err)
 	})
 
-	// drain with an inexisting Process
-	It("Honors grace and drain periods with stale PID", func() {
+	// drain with a non-existent Process
+	It("Stale PID should be ignored", func() {
 		haproxyBackendPort := 12000
 		// Expect initial deployment to be failing due to lack of healthy backends
 		haproxyInfo, _ := deployHAProxy(baseManifestVars{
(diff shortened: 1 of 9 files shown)


jobs/haproxy/templates/drain.erb
```diff
@@ -39,12 +47,24 @@ haproxy_master_pid=$(pgrep -P "$pid" -x haproxy)
...
 kill -USR1 "${haproxy_master_pid}"
 echo "$(date): triggering drain for process ${haproxy_master_pid}" >> ${logfile}
 
-echo <%= p("ha_proxy.drain_timeout") -%>
+drain_time=0
+drain_timeout=<%= p("ha_proxy.drain_timeout") %>
+while kill -0 "${haproxy_master_pid}" 2>/dev/null; do
+  sleep 1;
+  drain_time=$((drain_time + 1))
+  if [ $drain_time -eq $drain_timeout ]; then
```
peanball, line 62: I would recommend >= to avoid any possible "skipping" issues if this code were ever to be modified incorrectly.
```suggestion
  if [ $drain_time -ge $drain_timeout ]; then
```

jobs/haproxy/templates/haproxy_wrapper.erb
```diff
@@ -6,6 +6,7 @@ set -e
 export PATH=$PATH:/var/vcap/packages/haproxy/bin:/var/vcap/packages/ttar/bin
 CONFIG=/var/vcap/jobs/haproxy/config/haproxy.config
 PID_FILE=/var/vcap/sys/run/haproxy/haproxy.pid
+lockfile=/var/vcap/sys/run/haproxy/drain.lock
```
peanball, line 9:
```suggestion
DRAIN_LOCK=/var/vcap/sys/run/haproxy/drain.lock
```
to make the purpose of the variable clearer

jobs/haproxy/templates/haproxy_wrapper.erb
```diff
@@ -106,7 +107,5 @@ trap reload_daemon USR2
 echo "$(date): Starting HAProxy"
 haproxy -f ${CONFIG} -W -p ${PID_FILE} <%= flags.join(' ') %> <%= cat_pipe %>
 
-<%- if !run_in_foreground -%>
-# wait for the HAProxy main process to terminate
-while kill -0 $(cat ${PID_FILE}) 2>/dev/null; do sleep 1; done;
-<%- end -%>
+# wait for the HAProxy main process to terminate non-gracefully (drain)
+while [ -f ${lockfile} ] || kill -0 $(cat ${PID_FILE}) 2>/dev/null; do sleep 1; done;
```
peanball, line 111:
```suggestion
while [ -f ${DRAIN_LOCK} ] || kill -0 $(cat ${PID_FILE}) 2>/dev/null; do sleep 1; done;
```
see above.

peanball (changes requested):
* suggestion for making the lock "fail-safe" in case anything would go wrong with it
* Minor changes in the naming to make the purpose of variables clearer

acceptance-tests/drain_test.go
```diff
@@ -126,4 +135,68 @@ var _ = Describe("Drain Test", func() {
...
+			}, "\r\n")))
+
+			Expect(err).ToNot(HaveOccurred())
+
+			// Too lazy to properly parse headers? Just stop reading after a second!
+			_ = conn.SetReadDeadline(time.Now().Add(time.Second))
+
+			response, _ := io.ReadAll(conn)
+			return string(response)
+		}
```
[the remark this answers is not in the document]
domdom82: The issue is that the Go HTTP client is smart enough to filter out things like "connection: close" and act upon them (e.g. close the connection). Plus it is relatively hard to get from the HTTP client down to the underlying TCP connection again and see if it is actually closed or not. 
So, to avoid these problems I chose to use a low-level approach to just compare the text output of the HTTP response and see if it contains the "connection: close" header.

jobs/haproxy/templates/drain.erb
```diff
@@ -6,13 +6,21 @@ set -e
 pidfile=/var/vcap/sys/run/bpm/haproxy/haproxy.pid
 sockfile=/var/vcap/sys/run/haproxy/stats.sock
 logfile=/var/vcap/sys/log/haproxy/drain.log
+lockfile=/var/vcap/sys/run/haproxy/drain.lock
+
+mkdir -p "$(dirname ${logfile})"
```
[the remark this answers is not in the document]
domdom82: this was actually unrelated to the lock file but I noticed that in the template, in case drain is disabled, we actually write to the log file before the `mkdir` call, so it could potentially fail. So, to correct that I've moved the `mkdir` before the first log statement. (I actually think the `mkdir` may be obsolete because BOSH should create the `/var/vcap/sys/log/haproxy/` directory regardless, however I didn't verify that).

jobs/haproxy/templates/drain.erb
```diff
@@ -6,13 +6,21 @@ set -e
...
 echo 0
 exit 0
 <% else -%>
-mkdir -p "$(dirname ${logfile})"
+
+if [[ -f ${lockfile} ]]; then
+  echo "$(date): draining already in progress" >> ${logfile}
+  echo 0
+  exit 0
+fi
```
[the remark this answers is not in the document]
domdom82: the lock exists to decouple draining from restarts. the file will be removed during `haproxy_wrapper` startup in the `cleanup_daemon` function together with the rest (haproxy stats socket etc.)
we could think about adding a timestamp check here, but it's hard to find the right time out value:
- the time for draining will be at least grace_time + drain_time
- then there will be a potentially indefinite time of pre-stop scripts running
- then we have a potential bpm timeout of 15s to actually stop haproxy_wrapper
- then we have the pre-start script where we (potentially) wait for additional external certificates before we actually start
- finally, the time before haproxy has come online again, healthy and with the socket ready to accept commands, until we potentially can start a new drain again

The only possible way right now, this could get stuck is if the wrapper_script fails to start and fails to remove the lock file again. If the drain script starts draining but fails somewhere in-between, the next try will simply skip draining because the lock is there and stop haproxy without draining, clearing the file again and making it unstuck.

jobs/haproxy/templates/drain.erb
domdom82: agreed.

peanball (approved):

PR #336, opened 2022-11-30
Close connection on health checks

domdom82:
This PR fixes a bug with the Azure load balancer that prevents it from recognizing this HAproxy instance as unhealthy during draining.
The setting forces a "connection: close" from HAproxy as well as a FIN sent to the client. This forces the load balancer to open a new connection on every health check, which won't be possible during draining, because the listener will have been disabled by then.
The result is that the Azure load balancer immediately acknowledges this HAproxy as unhealthy and will stop sending new connections to it.

### DIFF
--- a/acceptance-tests/healthcheck_test.go
+++ b/acceptance-tests/healthcheck_test.go
@@ -2,7 +2,10 @@ package acceptance_tests
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


peanball (approved):
LGTM

PR #342, opened 2022-12-17
Use download_url instead of url for assets

### LINKED ISSUE #340: Autobumper for GitHub based projects seems to consider tags that don't have a release build yet
Example: #339 

There is a release in tar.gz form for 10.41 on the releases page, which is different from the tag tarball for `pcre2-10.41`.

domdom82:
Fixes #340


peanball (approved):
LGTM

PR #344, opened 2022-12-19
Add filename_suffix to support hatop assets

domdom82:
This should fix the problems with hatop not having a `*.tar.gz` asset

we will probably also have to fix the packaging script for hatop as we no longer have a tar gz to unpack


peanball:
> we will probably also have to fix the packaging script for hatop as we no longer have a tar gz to unpack
Could you add that to this PR or do you want to try it from master first and adjust?

peanball (approved):
Approving with the expectation that it will fail for hatop, which will need another change.

PR #346, opened 2023-01-02
Add `allow_all` property to allow TCP traffic explicitly

peanball:
This is useful when no allowlist is set, but all traffic should be allowed implicitly.


peanball:
Not the approach taken.

PR #352, opened 2023-01-24
Move request rate limit counting from TCP to HTTP level

peanball:


peanball:
Fully tested on aws-cfn02 (also disabling the allowlist, which leads to the tcp-level accept that was causing the requests not to be counted.)

maxmoehl (approved):

PR #356, opened 2023-02-02
Adapt `shipit` task to GCP/GCS blobstore

peanball:

PR #357, opened 2023-02-03
Shipit: Fix incorrect YAML padding

peanball:
Yaml padding broke in #356 after addressing review comments and running into a bash limitation (no variable expansion in sequence literals)

### DIFF
--- a/ci/scripts/shipit
+++ b/ci/scripts/shipit
@@ -26,17 +26,17 @@ header() {
 : "${VERSION_FROM:?required}"
 : "${GCP_SERVICE_KEY:?required}"
 
-if [[ ! -f ${VERSION_FROM} ]]; then
+if [[ ! -f "${VERSION_FROM}" ]]; then
   echo >&2 "Version file (${VERSION_FROM}) not found.  Did you misconfigure Concourse?"
   exit 2
 fi
-VERSION=$(cat ${VERSION_FROM})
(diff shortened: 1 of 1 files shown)


ci/scripts/shipit
```diff
@@ -153,8 +158,8 @@ if [[ -z $(git config --global user.name) ]]; then
 fi
 
 pushd "${REPO_ROOT}"
-  for MANIFEST_PATH in $(ls manifests/*.yml); do
-    $DIR/update-manifest ${GITHUB_OWNER} ${RELEASE_NAME} ${VERSION} ${SHA1} ${MANIFEST_PATH}
+  for MANIFEST_PATH in manifests/*.yml; do
```
plowin, line 161: cool, didn't know that worked!

plowin (approved):
thx for the cleanup, LGTM!

ci/scripts/shipit
peanball: `shellcheck` told me I should not parse outputs of `ls` and instead use a glob :D

I didn't know 100% either.

Let's see if it works. I will put this into a modified pipeline that pulls this branch for the shipit task. I'd like to avoid 1h+ round trips + approvals to fix typos (like I made them yesterday)

plowin: worked on my machine, should be fine! :) 👌

PR #361, opened 2023-02-15
fix(autobumper): update to use the GCS bucket

peanball:
The move to GCS did not address the autobumper. This PR now does.

Also fix support for alternative branches for PRs (branch-off point and non-conflicting autobumper branch name)


peanball:
Tested locally, which led to creation of #360.

plowin (approved):
LGTM and very similar to https://github.com/cloudfoundry/haproxy-boshrelease/pull/356/files

PR #395, opened 2023-03-28
Add versioning guide

b1tamara:


ci/README.md
```diff
@@ -110,3 +110,22 @@ Note that you can use the `dir` parameter in `run` to define the working directo
...
+
+  Used for new HAProxy **minor** versions, as they could contain breaking changes.
+* **Minor Version** (x.*Y*.z) -- incremented if new, backwards compatible functionality is introduced to the public API
+  
+  Used when upgrading dependencies (e.g. PCRE, socat, etc.) or HAProxy **patch** versions.
+* **Patch Version** (x.y.*Z*) -- incremented if only backwards compatible bug fixes are introduced)
+  
+  Used for documentation updates, changes in the test suite or any updates in the testing frameworks (e.g. ginkgo).
+
+The `haproxy-boshrelease` also contains patches (see [haproxy-patches](../haproxy-patches)). The patched version is denoted by appending a hyphen and key word `patched`, like `11.17.4-patched`.
```
maxmoehl, line 128: This is in conflict with the SemVer spec which uses hyphens (`-`) for pre-releases

maxmoehl (changes requested):

ci/README.md
maxmoehl: In theory the `patched` notion could also be considered build metadata, but I'm not sure if this is confusing and it might be easier to just denote this by the name of the binary.

peanball: I would suggest `11.17.4+2.6.9-patched`. It is actually the HAProxy that is patched and `-` is valid in the build metadata.

b1tamara: Applied

maxmoehl (approved):

PR #414, opened 2023-04-04
Remove slack notifications from maintenance pipeline

b1tamara:


.github/dependabot.yml
maxmoehl, line 1: Was this an intended change?

b1tamara: Yes. We do not want that the PRs to master and maintenance will come at the same time (due to PR validations)

plowin (approved):

PR #334, opened 2022-10-25
Canned Bosh Director for Acceptance Tests

peanball:
Add script for creating a running bosh director in Docker that can be resurrected.

This is for running locally only so far.
Integration with the concourse pipeline will be done in a future activity.

PR #418, opened 2023-04-05
fix(ci): bump go in testflight image and fix spruce dependency

peanball:
The `bosh/main-bosh-docker` image is deprecated. It is instead moving to `bosh/docker-cpi:main`.

Spruce was pulled in from `apt.starkandwayne.com`, which is not reachable anymore. Instead pulling the binary from GitHub releases directly.

Also replaces Go from the `golang:1.20` image (currently 1.20.3)

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,14 +1,12 @@
 FROM bosh/main-bosh-docker
 
-ARG GINKGO_VERSION=v2.2.0
+ARG GINKGO_VERSION=v2.9.2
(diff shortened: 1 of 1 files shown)


peanball:
We noticed that `bosh/main-bosh-docker` is not maintained anymore. This might be the starting point for a PR to move to a `bosh/docker-cpi:main` based image.

maxmoehl:
See also 2396cfa3fdeac213902d40431482f8522437edf8 in #334

PR #419, opened 2023-04-05
update gcs bucket to use haproxy-boshrelease

Mrizwanshaik:


ci/pipeline-haproxy-maintenance.yml
```diff
@@ -449,7 +449,7 @@ resources:
   - name: blobstore
     type: gcs
     source:
-      bucket:   haproxy-boshrelease
+      bucket:   haproxy-boshrelease-maintenance
```
peanball, line 452: It should use the same bucket. The `haproxy-boshrelease-maintenance` bucket was only created for testing to not interfere with the live environment but should be deleted afterwards.

Mrizwanshaik: okay but for semver version we are still using `haproxy-boshrelease-maintenance` https://github.com/cloudfoundry/haproxy-boshrelease/blob/d1aac0c8b3b6bd924dc3a03e4642ddcc5c71df6f/ci/pipeline-haproxy-maintenance.yml#L432 here I guess it should be like this?
and I see there is a config file here: https://github.com/cloudfoundry/haproxy-boshrelease/blob/064285b35ec133d1fea60dc30aa5bf7baee6ac58/config/final.yml#L6 here we should use haproxy-boshrelease-maintenance? as it is called in `shipit` script here: https://github.com/cloudfoundry/haproxy-boshrelease/blob/064285b35ec133d1fea60dc30aa5bf7baee6ac58/ci/scripts/shipit#L49

peanball: That should have been adapted.

Mrizwanshaik: I have change the bucket back to use `haproxy-boshrelease`. Will cross check version file with @b1tamara next week.

Mrizwanshaik:
Its just a proposal after testing on fork repo, we will discuss this next week together with tamara.
Note: `haproxy-boshrelease-maintenance` pipeline is configured to use forked repo. we can upload to use it from `cloudfoundry/haproxy-boshrelease` master once we decide to merge it or not.

ci/pipeline-haproxy-maintenance.yml
```diff
@@ -430,7 +430,7 @@ resources:
     source :
       driver:          gcs
       bucket:          haproxy-boshrelease
-      key:             version-11.16
+      key:             Version-11.16
```
maxmoehl: shouldn't the name be `maintenance-version` or similar to avoid manual changes to the key once we move the maintenance version up?

b1tamara: discussed with @maxmoehl , @peanball , @Mrizwanshaik. As the key we will use schema: version-<major>, e.g. version-11, version-12.

b1tamara: @Mrizwanshaik Could you please adapt the key to `version-11`

PR #430, opened 2023-04-25
Enable core-dumps on HAProxy crashes

maxmoehl:


maxmoehl:
Currently blocked because the bpm container prevents us from updating the limits for processes in the container. The bpm configuration does not support changing the core-dump limit natively as described in [the documentation](https://github.com/cloudfoundry/bpm-release/blob/master/docs/config.md#limits-schema).

Currently HAProxy fails to start because the wrapper script fails with:

```
sudo: The "no new privileges" flag is set, which prevents sudo from running as root.
sudo: If sudo is running in a container, you may need to adjust the container configuration to disable the flag.
```

PR #440, opened 2023-05-03
Update ci base image

Mrizwanshaik:
- Update CI base image to docker-cpi
- Removed pinned ruby vesrion `.ruby-version` and added ruby `~>3.2` version to gemfile


ci/pipeline-haproxy-maintenance.yml
```diff
@@ -30,7 +30,7 @@ jobs:
             type: docker-image
             source:
               repository: cf-routing.common.repositories.cloud.sap/haproxy-boshrelease-testflight
-              tag:        latest
+              tag:        main
```
peanball, line 33: I would keep the tag as "latest"

ci/scripts/unit-tests
```diff
@@ -4,5 +4,5 @@ set -euo pipefail
 
 cd ${REPO_ROOT}
 
-bundle package
+bundle update
```
peanball, line 7: Shouldn't this be bundle install, so it installs the versions from the gemfile and not what is the most current available?

spec/haproxy/templates/haproxy_config/backend_http_spec.rb
```diff
@@ -8,23 +8,23 @@
   end
 
   let(:properties) { {} }
-  let(:backend_http1) { haproxy_conf['backend http-routers-http1'] }
-  let(:backend_http2) { haproxy_conf['backend http-routers-http2'] }
+  let(:backend_http_one) { haproxy_conf['backend http-routers-http1'] }
```
peanball, line 11: Was this mandated by a linter? Not all linter rules are useful. We may choose to instead ignore this rule instead of sweeping name changes.

peanball (commented):
Some questions that need clarification.

ci/scripts/unit-tests
b1tamara: Yes we have to use `bundle install.` See help for install vs update:
For update: `Update the gems specified (all gems, if --all flag is used), ignoring the previously installed gems specified in the Gemfile.lock. In general, you should use bundle install(1) bundle-install.1.html to install the same exact gems and versions across machines.` But we want to install pinned versions.

peanball (approved):

PR #475, opened 2023-06-15
fix(ci): only trigger ci on approved PRs

domdom82:
- CI can now only be triggered with both "run-ci" label and a fresh approval by one of the code owners
- approvals will be revoked on code changes
- tested successfully on https://github.com/cloudfoundry/haproxy-boshrelease/pull/474
- pipeline already patched


maxmoehl:
@domdom82 can an approver still run the pipeline on a change that they'd like to propose without having to get an approval from someone else?

domdom82:
yes, the other alternative proposal was to disable forks. I don't like the current situation either. it's very restrictive now. we just now squashed the security exposure, let's discuss internally first how to proceed.

PR #493, opened 2023-07-25
Test PR for new ci-resource, please ignore

plowin:


plowin:
Works as designed, PRs from @cloudfoundry/wg-app-runtime-platform-networking-extensions-approvers, dependabot and the CFN-CI user will be run on the CI when label `run-ci` is set.
Others PRs need a review by a WG approver and the label

PR #507, opened 2023-08-15
ci(shipit): use a bot account for creating release commits

peanball:
The `shipit` script creates commits as part of the release process and pushes them to the `master` branch.

The author of the commit must be excluded from the branch protection from the master branch. This is only the case for our bot users.

This change allows setting the git username / e-mail to the botuser (maintained in the vars.yml for the pipeline)


plowin (approved):
lgtm!

PR #536, opened 2023-10-10
Make it possible to enable Haproxy Metrics endpoint

benjaminguttmann-avtq:
This PR should make it possible to enable Haproxy's metrics endpoint.
I think the .erb needs a bit more adjustments to untangle stats/metrics properties but in general it is working already like that


maxmoehl (commented):
Already looking good. It probably doesn't hurt to always compile with prometheus, the difference seems to be negligible:
```
-rwxrwxr-x  1 ubuntu ubuntu 15562848 Oct 10 17:55 hap
-rwxrwxr-x  1 ubuntu ubuntu 15700216 Oct 10 17:57 hap_promex
```

Since it's still WIP I won't comment on the missing tests :smile:

Regarding the properties: we should probably put them next to the stats properties and align the naming as we re-use the stats frontend (and surrounding functionality like the ACL).

benjaminguttmann-avtq:
@maxmoehl Thanks for adding feedback that fast (and mentioning the tests). Will try to figure out how to add the tests :)

maxmoehl:
> Regarding the properties: we should probably put them next to the stats properties and align the naming as we re-use the stats frontend (and surrounding functionality like the ACL).

To elaborate a bit more: if you enable the prometheus endpoint but do not enable the stats endpoint you won't get the prometheus endpoint. Probably a logic like this is more desirable:

```
IF prom OR stats:
  add frontend stats

IF prom:
  configure prom
IF stats:
  configure stats
```

Any other opinions on this? 😅

peanball:
@benjaminguttmann-avtq, from my point of view Prometheus is still 'stats'.

You configure stats with a port (default port if not mentioned) and enable HAProxy CSV stats, Prometheus or both.

I wouldn't see Prometheus as some special thing, as HAProxy doesn't either.

maxmoehl:
> @benjaminguttmann-avtq, from my point of view Prometheus is still 'stats'.
> 
> You configure stats with a port (default port if not mentioned) and enable HAProxy CSV stats, Prometheus or both.
> 
> I wouldn't see Prometheus as some special thing, as HAProxy doesn't either.

@peanball so just put prometheus in the `ha_proxy.stats_` namespace and configure it as one thing (e.g. `ha_proxy.stats_promex_enable` and `ha_proxy.stats_promex_path`)?

Sounds good to me.

benjaminguttmann-avtq:
@maxmoehl adjusted accordingly + added tests

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -298,6 +298,10 @@ listen stats
     acl private src <%= p("ha_proxy.trusted_stats_cidrs") %>
     http-request deny unless private
     mode http
+  <%- if_p("ha_proxy.stats_promex_enable") do -%>
+    http-request use-service prometheus-exporter if { path <%= p("ha_proxy.stats_promex_path") %> }
+    option http-use-htx
```
maxmoehl, line 303: This config option no longer exists as of haproxy/haproxy@9862787e8f4de3a901717565a6828fa3288919ba.
```suggestion
```

benjaminguttmann-avtq:
@maxmoehl removed the option and adjusted the tests as well

maxmoehl (approved):
Let's see what the CI has to say :smile:

spec/.ruby-version
maxmoehl, line 1: Do you know where this file came from?

benjaminguttmann-avtq: yes from me not planning to add that one :D

maxmoehl (approved):
CI.

PR #544, opened 2023-10-20
refactor: fix parameter order in create_pull according to PyGithub 2.1.X

b1tamara:
Due to bump of pygithub from 1.59.1 to 2.1.1, the call of create_pull should be adapted. See breaking change in release notes for pygithub for more reference. 

new function signature for create_pull:
`def create_pull(self, base, head, *, title=NotSet, body=NotSet, maintainer_can_modify=NotSet, draft=NotSet, issue=NotSet)`

### DIFF
--- a/ci/scripts/autobump-dependencies.py
+++ b/ci/scripts/autobump-dependencies.py
@@ -208,10 +208,11 @@ class Dependency:
             )
 
             pr = self.remote_repo.create_pull(
-                title=f"Bump {self.name} version to {self.latest_release.version}",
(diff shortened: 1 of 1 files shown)


maxmoehl:
Has this caused any errors? Since we use named parameters the order shouldn't matter.

b1tamara:
It is related to https://github.com/cloudfoundry/haproxy-boshrelease/pull/530 which has been just merged.

b1tamara:
Close the PR as change is not needed.

PR #551, opened 2023-10-25
ci: update version file for releases

b1tamara:


haproxy-boshrelease.tgz
maxmoehl, line 1: Please be careful about what you check in!

maxmoehl (changes requested):

haproxy-boshrelease.tgz
b1tamara: oh good point

maxmoehl (approved):

PR #583, opened 2023-12-18
feat: Enable redispatch to different server on retry

domdom82:
Without retries and redispatch enabled, failing backends produce an outage to the client.
The redispatch option allows HAProxy to try another backend instead of failing the request.

- Adds the option to retry failed connections to backends
- Adds the option to redispatch to other backends on failed connects
- Checks that retries are configured if redispatch is enabled

- Also, the test framework was updated so that it supports more than one backend per HAProxy.

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -172,6 +172,10 @@ end
     abort "Conflicting configuration: if enable_4443 is true, you must provide a valid SSL config via ssl_pem or crt_list"
   end
 
+  if p("ha_proxy.retries") == 0 && p("ha_proxy.enable_redispatch")
+    abort "Conflicting configuration: enable_redispatch works only with retries > 0"
+  end
+
   backend_servers = []
   backend_servers_local = []
   backend_port = nil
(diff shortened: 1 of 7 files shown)


jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -281,6 +285,11 @@ defaults
   <%- if p("ha_proxy.backend_prefer_local_az") -%>
     option allbackups
   <%- end -%>
+  <%- if p("ha_proxy.enable_redispatch") -%>
+    option redispatch
+  <%- end -%>
+    retries <%= p("ha_proxy.retries") %>
```
maxmoehl, line 291: Maybe move this into the if as well? I'd prefer if we keep the configuration as small as possible.

domdom82: they are two distinct properties. you can have retries without redispatch. but I can add it only if the value is larger than 0

maxmoehl: Ah, got it. In that case I don't have a preference.

maxmoehl (approved):

PR #601, opened 2024-01-17
ci: de-couple unit and acceptance tests

maxmoehl:
Concourse is not the best when it comes to PR validations. To reduce the amount of stuck builds and PRs this commit de-couples the two checks. This might result in more wasted builds on the acceptance tests which are expensive, but the current situation which requires manually triggering the checks is much worse.


maxmoehl:
I've pushed the pipelines, let's see if it fixes the issue and we can review and merge in the next days.

domdom82 (approved):

PR #646, opened 2024-04-04
Remove maintenance build

domdom82:
Maintenance pipeline already removed from concourse.

Maintenance branch to be removed after this PR is merged.


peanball:
Could you add a pointer in the docs (e.g. readme) that we had a maintenance branch, that it is disabled and in which commit to find the alternative pipeline?

domdom82:
> Could you add a pointer in the docs (e.g. readme) that we had a maintenance branch, that it is disabled and in which commit to find the alternative pipeline?

Well, the commit is this PR, right? All we have to do is revert it and we got the pipeline back. I'll add it to the readme though.

PR #651, opened 2024-04-17
ci: add release notes

maxmoehl:


plowin (approved):
also many CI enhancements, could be mentioned in the rel-notes but can also be spared.

PR #655, opened 2024-04-26
ssl_min/max_ver, pre_start_script, config_mode raw_blocks_only, raw_blocks without names

kinjelom:
1. `ssl_min_ver` and `ssl_max_ver`: These settings enforce minimum and maximum SSL versions for connections. If this option is set, the `disable_tls_*` options will be ignored.

2. `pre_start_script`: A script appended to the pre-start script of HAProxy job. This script allows the job to prepare the machine before starting HAProxy, for example, by setting the MTU to a custom value.

3. `config_mode`: This includes two modes:
   - 'auto': default, utilizes raw_config if defined; otherwise, it uses traditional configuration mixed with raw_blocks.
   - 'raw_blocks_only': uses only raw_blocks, ignoring other configurations.

4. support for `raw_blocks` with or without names, this is more readable for blocks global and defaults, especially when applying the setting `config_mode=raw_blocks_only`.


kinjelom:
I'll clean up my repo and do it again

PR #657, opened 2024-04-29
ssl_min/max_ver, pre_start_script, raw_blocks without names and raw_blocks_only config_mode, stats without authentication

kinjelom:
1. `pre_start_script`: A script appended to the pre-start script of HAProxy job. This script allows the job to prepare the machine before starting HAProxy, for example, by setting the MTU to a custom value.

2. `ssl_min_ver` and `ssl_max_ver`: These settings enforce minimum and maximum SSL versions for connections. If this option is set, the `disable_tls_*` options will be ignored.

3. support for `raw_blocks` with or without names, this is more readable for blocks global and defaults, especially when applying the setting `config_mode: raw_blocks_only`.
   `config_mode` includes two modes:
   - `auto`: default, utilizes raw_config if defined; otherwise, it uses traditional configuration mixed with raw_blocks.
   - `raw_blocks_only`: uses only raw_blocks, ignoring other configurations.

4. Ability to expose `ha_proxy.stats` without authentication.


plowin:
Hi @kinjelom, can you please resolve the conflict, a typo fix? I do not want to push to your fork's `master` branch.

kinjelom:
> Hi @kinjelom, can you please resolve the conflict, a typo fix? I do not want to push to your fork's `master` branch.

Sorry, I missed it, done.

jobs/haproxy/spec
```diff
@@ -237,21 +259,32 @@ properties:
...
+      the 'disable_tls_*' options. If this option is set, the 'disable_tls_*' options will be ignored.
+  ha_proxy.ssl_max_ver:
+    example: "TLSv1.3"
+    description: |
+      This option enforces the use of 'version' or lower for SSL connections initiated from this listener.
+      It will only be set if 'ssl_min_ver' is specified, as the default HAProxy ssl-min-ver may change in future.
   ha_proxy.disable_tls_tickets:
     default: true
-    description: "Improve (Perfect) Forward Secrecy by disabling TLS tickets"
+    description: "Improve (Perfect) Forward Secrecy by disabling TLS tickets. Use 'ssl_min_ver' and 'ssl_max_ver' instead."
```
plowin, line 275: I think we need to keep the option for tls_tickets also with the new style, as people could configure a lower min-version for some legacy system. It does not seem deprecated in the [docs](https://docs.haproxy.org/2.8/configuration.html#5.1-no-tls-tickets)

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -40,21 +41,32 @@ if p("ha_proxy.disable_tcp_accept_proxy")
...
+  if p("ha_proxy.disable_tls_11")
+    ssl_flags = "#{ssl_flags} no-tlsv11"
+  end
+  if p("ha_proxy.disable_tls_12")
+    ssl_flags = "#{ssl_flags} no-tlsv12"
+  end
+  if p("ha_proxy.disable_tls_13")
+    ssl_flags = "#{ssl_flags} no-tlsv13"
+  end
+  if p("ha_proxy.disable_tls_tickets")
```
plowin, line 67: Move this one level higher if you agree to the comment in the spec

jobs/haproxy/spec
```diff
@@ -39,6 +39,28 @@ consumes:
...
+      CUST_MTU=((custom_mtu))
+      INTERFACE=$(ip -4 route get 8.8.8.8 | grep -Po '(?<=dev )\S+') #'
+      CURR_MTU=$(ip link show $INTERFACE | grep -Po 'mtu \K\d+')
+      if [[ $CURR_MTU -ne $CUST_MTU ]]; then
+        sudo ip link set dev $INTERFACE mtu $CUST_MTU
+        echo "MTU changed from $CURR_MTU to $CUST_MTU, interface: $INTERFACE."
+      fi
+      CURR_MTU=$(ip link show $INTERFACE | grep -Po 'mtu \K\d+')
+      echo "MTU: $CURR_MTU, interface: $INTERFACE"
+  ha_proxy.config_mode:
```
plowin, line 59: Can you please elaborate a bit more on your use-case requiring this additional switch? Maybe you can provide your concrete example? The logic around these options is already not trivial and I am not sure if this would be backwards compatible for all users.

In general, I'd prefer a PR per feature to be flexible for fixes when unforseen problems are caused for some stakeholders. If we can ensure backwards compatibility, we can also make an exception for those mid-sized enhancements. I agree to the other three changes but struggle to wrap my head around this one.

plowin (changes requested):
Thanks a lot for the contribution, I added some minor remarks and questions.

jobs/haproxy/spec
kinjelom: You're right, I was convinced that I had read in the documentation that this option disappears with the transition to ssl-min/max-ver, however, that is not the case. https://docs.haproxy.org/2.9/configuration.html#5.1-no-tls-tickets

jobs/haproxy/spec
kinjelom: At the moment, we have an automatic configuration mode, but it's implicit, which is why the default value is `auto`. `auto` means `utilizes raw_config if defined; otherwise, it uses traditional configuration mixed with raw_blocks`.

I would also like the option to use only `raw_blocks` as a more readable and flexible version of `raw_config`. `raw_blocks` enhance operations through bosh-ops and provide clearer diffs.

Currently, this can be achieved by inserting an empty configuration into `raw_config`, but this is not a good or clear solution.
It seems to me that clarifying the configuration mode is more understandable, and certainly, the default value `auto` is backward compatible.

plowin: Okay, I can live with that. However, I think the current version of the spec is not fully correct. 
> 'auto' - utilizes raw_config if defined; otherwise, it uses traditional configuration mixed with raw_blocks
Should probably rather be
> 'auto' - utilizes raw_config if defined and mixes it with `raw_blocks`; otherwise, it uses traditional configuration mixed with `raw_blocks`

WDYT?

plowin (dismissed):
Approving to run the CI. Before merge, I'd prefer rebased commits from the current upstream master-branch. Please also squash the recent adjustments with a proper commit-message

jobs/haproxy/spec
kinjelom: Now the template is as follows:
```
if properties.ha_proxy.config_mode == "auto"
  if properties.ha_proxy.raw_config
    # ha_proxy.raw_config
  else
    # traditional/static config
  end
end

if properties.ha_proxy.raw_blocks && !properties.ha_proxy.raw_blocks.empty? -%>
  # raw blocks
end
```
Therefore, if the `config_mode` is `auto`, it will be either `raw_config` (if not empty) or static mixed with `raw_blocks`.
If the `config_mode` is other than auto (currently this can only be raw_blocks_only), then the configuration will only contain `raw_blocks`, provided they are defined.

You are right. If the `config_mode` is auto, and both `raw_config` is not empty and `raw_blocks` are defined, there could be a problem...
So, the template should be:
```
if properties.ha_proxy.config_mode == "auto"
  if properties.ha_proxy.raw_config
    # ha_proxy.raw_config
  else
    # traditional/static config
  end
end

if !properties.ha_proxy.raw_config && properties.ha_proxy.raw_blocks && !properties.ha_proxy.raw_blocks.empty?
  # raw blocks
end
```

The `config_mode: raw_blocks_only` is not used directly, but it opens the code for future development.

plowin:
@kinjelom , when rebasing, please also include the findings from the linter:
```
spec/haproxy/templates/haproxy_config/raw_blocks_spec.rb:92:1: C: [Correctable] Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body end.
spec/haproxy/templates/haproxy_config/stats_listener_spec.rb:88:1: C: [Correctable] Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body end.
```

plowin (dismissed):
lgtm

kinjelom:
> @kinjelom , when rebasing, please also include the findings from the linter [...]

@plowin done, is everything okay now?

plowin (approved):
Thanks, triggering CI.

plowin:
You are very welcome, thanks a lot for the contributions, released with https://github.com/cloudfoundry/haproxy-boshrelease/releases/tag/v14.0.0%2B2.8.9.

Feel free to also join us on [Slack](https://cloudfoundry.slack.com/archives/C0XRT9L22)

PR #661, opened 2024-05-07
ci: use ssh key for CFN-CI bot in concourse pipeline

b1tamara:
Problem:
It is not possible for the CFN-CI bot to bypass the branch protection rules. This is due to the fact that this user is no admin on this repo here.

Solution
Make use of a deploy key added to this repo within the git-resources.


plowin (approved):
Nice, thx, lgtm!

PR #680, opened 2024-06-14
Auto-bump socat to 1.x.y

peanball:
Updates the auto-bumper for socat from pinned version 1.7.x to 1.x.

Also pulls up the various pinned versions to the top of the script where they are easier to find.

### DIFF
--- a/ci/scripts/autobump-dependencies.py
+++ b/ci/scripts/autobump-dependencies.py
@@ -20,7 +20,7 @@ from git import Repo
 
 # Pinned versions for the resolvers. See use at the bottom.
 KEEPALIVED_VERSION = "2.2"
-SOCAT_VERSION = "1.7"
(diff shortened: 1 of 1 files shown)


ci/scripts/autobump-dependencies.py
```diff
@@ -18,6 +18,13 @@
 from bs4 import BeautifulSoup
 from git import Repo
 
+# Pinned versions for the resolvers. See use at the bottom.
+KEEPALIVED_VERSION = "2.2"
+SOCAT_VERSION = "1.8"
```
maxmoehl, line 23: Do we need to pin to 1.8? Wouldn't 1 be sufficient?

peanball: fair point.

maxmoehl (approved):

PR #689, opened 2024-07-17
feat: Strip headers

b1tamara:
Implement a feature to remove specific list of headers unconditionally.


hoffmaen:
Successfully tested this change manually with a test application:
* Headers listed in `strip_headers` are removed
* Headers listed in `headers` and `strip_headers` are forwarded with the value from `headers`

PR #690, opened 2024-07-19
Add strip_header to delete headers on the frontend

hoffmaen:
Implement config property strip_headers, which allows to delete configured headers from incoming requests based on the header key.

If both 'strip_headers' and 'headers' contain the same key, the header (if set) is first deleted from the request, and then set to the configured value.

### DIFF
--- a/acceptance-tests/headers_test.go
+++ b/acceptance-tests/headers_test.go
@@ -1,0 +1,143 @@
+package acceptance_tests
+
+import (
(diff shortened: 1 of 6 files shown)


b1tamara (dismissed):
Approve to run ci

b1tamara (approved):
Approve to run CI

b1tamara (approved):
LGTM

PR #693, opened 2024-07-29
Autobump to newest keepalived minor 2.3.x

plowin:


maxmoehl (approved):

plowin:
For reference: in the team we discussed pinning the major version to avoid such changes in the future. But to have more visibility of minor bumps and to raise a sentence in the release notes, we decided to stick with the status quo

PR #699, opened 2024-08-06
Adding option to check backend health via TLS

schmidtsv:
Hello haproxy people,

CF Gorouter wants to switch to TLS healthchecks. I tested out if the HAProxy works with it. I found the issue that unless you tell it explicitly to only use TLS for health checks it will run into TLS issues (namely not using it). THis PR aims to add an option to enable check-ssl (https://docs.haproxy.org/2.4/configuration.html#5.2-check-ssl) to the release.

I tested it working by hand originally on our staging and made a PR to replicate the setup with the bosh release. The Linter and test suite are green and without complaints. I also added a simple test for the property.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -677,7 +677,9 @@ properties:
   ha_proxy.backend_health_rise:
     description: Number of consecutive successful health checks required before the server is considered healthy from an unhealthy state. The default value of 2 matches the default if the parameter is undefined. This parameter will be ignored if ha_proxy.backend_use_http_health is false.
     default: 2
-
+  ha_proxy.backend_https_check:
+    description: Set to true if the backend uses TLS on the health endpoint. Adds the check-ssl option to the backend configs. If backend certificate on traffic port is verified the Health endpoint cert will also be verified.  
+    default: false
   ha_proxy.global_config:
     description: |
       Raw HAProxy config that will be added to the HA proxy global section, provided either as a multiline text blob or as an array of lines.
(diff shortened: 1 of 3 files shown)


jobs/haproxy/spec
```diff
@@ -677,7 +677,9 @@ properties:
   ha_proxy.backend_health_rise:
     description: Number of consecutive successful health checks required before the server is considered healthy from an unhealthy state. The default value of 2 matches the default if the parameter is undefined. This parameter will be ignored if ha_proxy.backend_use_http_health is false.
     default: 2
-
+  ha_proxy.backend_https_check:
+    description: Set to true if the backend uses TLS on the health endpoint. Adds the check-ssl option to the backend configs. If backend certificate on traffic port is verified the Health endpoint cert will also be verified.  
+    default: false
```
maxmoehl, line 682: Could you please re-introduce an empty new-line after the health-check section?

maxmoehl (approved):
Just a minor nit-pick, I'm approving now to run the CI.

maxmoehl:
As a side-note: for our environments we decided to stick to non-TLS health-checks to reduce the amount of TLS handshakes which need to be performed and save some resources. But it's always good to offer the option for those that want it!

jobs/haproxy/spec
schmidtsv: Done

schmidtsv:
We are at evaluating the impact of TLS checks at the moment. Though generally its probably a good idea to offer it at least, even if most people won't use it right now. 

As once caveat you either need to resign the health cert for the Gorouter with the router-ca or tell haproxy both router and service_cf_internal_ca. (for whoever comes in 3 years debugging heir ssl error).

maxmoehl (approved):

PR #716, opened 2024-10-24
fix: docker startup in acceptance

mtekel:
Do not set -e, because then we trap service docker stop and docker never starts. Instead keep waiting.


maxmoehl (approved):
Approving for CI.

maxmoehl (approved):
Approving for CI.

b1tamara (approved):
Approving for CI

PR #724, opened 2024-11-06
Update CI information

mtekel:
Update upload script and README to reflect recent changes with concourse server.

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -1,8 +1,9 @@
 # Continuous Integration Setup
 
-The CI environment is maintained by the SAP Business Technology Platform on Cloud Foundry team for Routing and Networking.
+The CI environment is maintained by the SAP Business Technology Platform on Cloud Foundry [team for Routing and Networking](https://github.com/orgs/cloudfoundry/teams/wg-app-runtime-platform-networking-extensions-approvers).
 
-All secrets used in the pipeline and surrounding tools are kept in the team's Vault. Those secrets are only necessary for [modifying the pipeline](#updating-or-modifying-the-pipeline).
+All secrets used in the pipeline and surrounding tools are kept in the team's Vault. Those secrets are only necessary for [modifying the pipeline](#updating-or-modifying-the-pipeline). For git integration, [haproxy-boshrelease pipeline](https://concourse.arp.cloudfoundry.org/teams/main/pipelines/haproxy-boshrelease) has public visibility.
+
 ## Contact
 
 Credentials and admin access lie with that team. If you have questions or issues, please reach out via the [#haproxy-boshrelease](https://cloudfoundry.slack.com/archives/C0XRT9L22) channel on Slack.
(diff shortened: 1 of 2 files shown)


ci/upload-to-concourse.sh
```diff
@@ -2,35 +2,16 @@
...
     exit 1
 fi
 
+echo "Please follow https://github.com/cloudfoundry/routing-concourse/blob/main/README.md#authentication-to-concourse for login."
 
-fly -t "$CONCOURSE_TARGET" login -c "$CONCOURSE_URL" -u "$CONCOURSE_USER" -p "$CONCOURSE_PASSWORD"
-fly -t "$CONCOURSE_TARGET" validate-pipeline -c pipeline.yml
-fly -t "$CONCOURSE_TARGET" set-pipeline -p haproxy-boshrelease -c pipeline.yml --load-vars-from vars.yml
-fly -t "$CONCOURSE_TARGET" expose-pipeline -p haproxy-boshrelease
+fly -t cfi login -c https://concourse.arp.cloudfoundry.org/
```
maxmoehl, line 12: Would it still be possible to change the team name? CFI is an outdated reference to an SAP internal team name, something like `networking-extensions` would be more appropriate.

ci/README.md
```diff
@@ -1,8 +1,9 @@
 # Continuous Integration Setup
 
-The CI environment is maintained by the SAP Business Technology Platform on Cloud Foundry team for Routing and Networking.
+The CI environment is maintained by the SAP Business Technology Platform on Cloud Foundry [team for Routing and Networking](https://github.com/orgs/cloudfoundry/teams/wg-app-runtime-platform-networking-extensions-approvers).
+
+All secrets used in the pipeline and surrounding tools are kept in the team's Vault. Those secrets are only necessary for [modifying the pipeline](#updating-or-modifying-the-pipeline). For git integration, [haproxy-boshrelease pipeline](https://concourse.arp.cloudfoundry.org/teams/main/pipelines/haproxy-boshrelease) has public visibility.
```
maxmoehl, line 5: This means that SAP employees are still the only ones who can change the pipelines?

ci/upload-to-concourse.sh
mtekel: It can be any name, I have re-used what we mention in other docs: https://pages.github.tools.sap/cloudfoundry/documentation/internal-documentation/cfn-haproxy-boshrelease-os#hijack-latest-concourse-jobs
I use `concourse-arp` internally. It can be anything, but needs updating in both places (here and any other docs).

ci/README.md
mtekel: Let's say vault provides convenience. Otherwise you can get credentials from individual places (e.g. if you need slack webook that's used for reporting build failures, probably you can get it from slack directly - but you need to have correct level of access). I think some people from cloudfoundry org have access to all the individual places and credentials that are needed.

ci/upload-to-concourse.sh
maxmoehl: Could you then please rename it to `networking-extensions` as the pipelines are from that WG area? It will also allow us to separate the other pipelines by area to give proper permissions to reviewers / approvers of the respective area.

<sub>And please try to keep references to SAP internal resources out of GitHub.com</sub>

maxmoehl (dismissed):

maxmoehl (approved):

ci/upload-to-concourse.sh
peanball: or `arp`?

ci/README.md
peanball: I think for the scope of this task we could leave it like that. It would be feasible to move those (community) secrets to some secret store on GCP, I guess?

PR #733, opened 2024-11-22
Periodically build and use testflight image

mtekel:
Build new image when there is new docker-cpi image, or when we change image config.
Build daily to bring updated os dependencies.
Use new image in the pipeline.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -13,6 +13,7 @@ groups:
       - minor
       - major
       - autobump-dependencies
+      - build-haproxy-testflight-image
 
 jobs:
   - name: unit-tests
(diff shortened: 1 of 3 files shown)


ci/pipeline.yml
```diff
@@ -379,10 +336,39 @@ jobs:
...
+        image: docker-cpi-image
+        config:
+          outputs:
+            - name: tag
+          platform: linux
+          run:
+            path: sh
+            args:
+            - -exc
+            - date +%Y%m%d.%H%M%S | tee tag/today
```
maxmoehl, line 363: This would result in `20241125.081050`, right? Can we make the date a bit easier to read, e.g. `2024.11.25-081050`?

maxmoehl (approved):

PR #735, opened 2024-11-28
ci: Use cfcommunity/github-pr-resource

mtekel:
Instead of depending on custom image, use standard cf community github pr resource.


peanball:
I added a sentence to the PR description that the custom fork is not needed anymore.

PR #736, opened 2024-12-03
Fix Draining for TCP and PROXY

maxmoehl:


peanball (approved):

maxmoehl:
This is still lacking tests, I'm working on it.

maxmoehl:
The TCP part is a lot more work and I don't have the time to work on it right now. The broken PROXY protocol is an actual bug and I've opened #737 to address it. This will have to wait.

PR #754, opened 2025-01-08
Fix acceptance test setup to work on local linux

mtekel:
Few minor fixes to acceptance test scripts and setup so that they can run fine on localhost running linux.

### DIFF
--- a/ci/scripts/acceptance-tests
+++ b/ci/scripts/acceptance-tests
@@ -16,6 +16,8 @@ fi
 
 cd "${REPO_ROOT:?required}"
 echo "----- Pulling in any git submodules..."
+git config --global --add safe.directory /repo
+git config --global --add safe.directory /repo/src/ttar
 git submodule update --init --recursive --force
 
 # shellcheck disable=SC1091
--- a/ci/scripts/shell
+++ b/ci/scripts/shell
@@ -7,6 +7,8 @@ bpm_release_path=$PWD/bpm/*.tgz
 
 cd ${REPO_ROOT:?required}
 echo "----- Pulling in any git submodules..."
+git config --global --add safe.directory /repo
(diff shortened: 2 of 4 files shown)


ci/scripts/acceptance-tests
```diff
@@ -16,6 +16,8 @@ fi
 
 cd "${REPO_ROOT:?required}"
 echo "----- Pulling in any git submodules..."
+git config --global --add safe.directory /repo
+git config --global --add safe.directory /repo/src/ttar
```
Dariquest, line 20: It make senses to introduce a variable for the path /repo/src/ttar, since you use it in another scripts in the same folder (ci/scripts)

ci/scripts/shell
```diff
@@ -7,6 +7,8 @@ bpm_release_path=$PWD/bpm/*.tgz
 
 cd ${REPO_ROOT:?required}
 echo "----- Pulling in any git submodules..."
+git config --global --add safe.directory /repo
+git config --global --add safe.directory /repo/src/ttar
```
Dariquest, line 11: It make senses to introduce a variable for the path /repo/src/ttar, since you use it in another scripts in the same folder (ci/scripts)

Dariquest (changes requested):

mtekel:
This is more of patching, just to make it work. Scripts run with -e, so any issue in the middle of the execution (which can take say 10 or 15 mins to arrive to that point) means aborting of everything.

There also is a lot of duplication (in general) between these two scripts (`run-local.sh` and `shell` script in `ci` folder). Then there's handling of BPM release (also being fixed in this PR) - but haproxy brings own version and uses that instead. 

We have decided to create improvements/tech debt story to address all of that. This PR remains "light" to get it working for now.

maxmoehl (approved):
Approving for CI.

PR #762, opened 2025-02-06
ci: adjust codeowners

maxmoehl:


maxmoehl:
This will need for [peribolos](https://github.com/cloudfoundry/community/actions/runs/13181825703) to do its sync and then we will see who has to approve the PR to get it in.

PR #759, opened 2025-02-04
Add parameter to set a header containing the client ip

hoffmaen:
Add property `ha_proxy.true_client_ip_header`. This property allows to define a header name that stores the client [src](https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#src) IP address, as seen by HAProxy.

PR #778, opened 2025-03-27
Retain True-Client-Ip for Route Services

a18e:
This Pull-Request completes the True-Client-Ip feature introduced with #759. With this change, we add three options to deal with the header: 
* `forward_only_if_route_service`: Forward only in route service scenario. This is the default behavior which adds support for the True-Client-Ip header in the route service scenario.
* `always_set`: Always set the header, overwrite if existing.
* `always_forward`: Always forward if existing, set otherwise.

### DIFF
--- a/acceptance-tests/headers_test.go
+++ b/acceptance-tests/headers_test.go
@@ -3,7 +3,6 @@ package acceptance_tests
 import (
 	"crypto/tls"
 	"fmt"
-	"net"
 	"net/http"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -20,9 +19,6 @@ var _ = Describe("Headers", func() {
   value: 
     Custom-Header-To-Add: add-value
     Custom-Header-To-Replace: replace-value
-- type: replace
-  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/true_client_ip_header?
-  value: "X-CF-True-Client-IP"
 # Configure CA and cert chain
 - type: replace
   path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/crt_list?/-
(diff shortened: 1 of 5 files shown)


b1tamara (dismissed):
approving to run CI

mtekel (approved):

mtekel (approved):

jobs/haproxy/spec
```diff
@@ -406,7 +406,21 @@ properties:
...
+      Your options are:
+      
+      - always_forward: Always preserve the existing header.
+      
+      - forward_only_if_route_service: This option is useful to support the header with Route Services.
+          The header will be overwritten with the current source address, unless there is an X-Cf-Proxy-Signature header.
+      
+      - always_set: Always set the header with the current source address. Will overwrite any existing header.
+
+    default: forward_only_if_route_service
```
maxmoehl, line 423: I know we discussed this offline, but I forgot what the reason was to have a breaking change in this property by changing the default behaviour?

acceptance-tests/headers_test.go
```diff
@@ -164,9 +164,39 @@ var _ = Describe("Headers", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Correctly preserves the pre-existing True-Client-Ip Header")
+		resp, err := client.Do(request)
+		expect200(resp, err)
+		headerKey := "X-Cf-True-Client-Ip"
+		Expect(recordedHeaders).To(HaveKey(headerKey))
+		Expect(recordedHeaders[headerKey]).To(HaveLen(1))
+		Expect(recordedHeaders[headerKey][0]).To(Equal("8.8.8.8"))
+	})
```
maxmoehl, line 199: Please add tests for the other two cases as well.

jobs/haproxy/spec
a18e: Because it's a sane default. We argued that the feature wasn't properly implemented because it just didn't work for route services at all. Now it's unified/harmonized by default.

b1tamara (approved):
Approving to run CI

maxmoehl (approved):
Once merged, please cut a major release right away to make sure we don't accidentally release this with a minor / patch.

PR #782, opened 2025-04-01
Improve acceptance tests

mtekel:
Fix tech debt on acceptance tests:
- remove unused bpm

Improve runtime/user friendliness for development and debug:
- use local/cached testflight image
- DRY wrapper scripts
- run with "dirty" git
- add persistence for containers (bosh stays running)
- optimise speed (compilation = 2, working overlay2 in bosh docker for internal docker)
- apply shellcheck recommended fixes

Update docs.

### DIFF
--- a/acceptance-tests/README.md
+++ b/acceptance-tests/README.md
@@ -3,10 +3,7 @@
 ## Requirements
 
 * Docker installed locally
-* A matching Jammy stemcell tgz downloaded to `ci/scripts/stemcell`
-  * Get it from https://bosh.io/stemcells/bosh-warden-boshlite-ubuntu-jammy-go_agent
-* A matching Bionic stemcell tgz downloaded to `ci/scripts/stemcell-bionic`
-  * Get it from https://bosh.io/stemcells/bosh-warden-boshlite-ubuntu-bionic-go_agent
+* x86 based computer (cloudfoundry currently doesn't build/run on arm; rosetta can't run docker in docker for x86)
 
 ## Running
 
@@ -16,6 +13,8 @@ cd acceptance-tests
 ```
 
 ### Running on Docker for Mac
+
+Unfortunately rosetta on arm based macs doesn't support docker in docker required for bosh docker-cpi used in this test setup. You will either have to run on x86 based mac or some remote x86 workstation. Virtualization using QUEMU is possible, but so slow the tests fail on timeouts.
 
 The BOSH Docker CPI requires cgroups v1 to be active. Docker for Mac since 4.3.x uses cgroups v2 by default.
 
@@ -80,24 +79,40 @@ The output at the end should be:
 
 If you want to run only a specific part of the suite, you can use [focussed specs](https://onsi.github.io/ginkgo/#focused-specs)
 
-The easiest way is to just provide the name of the tests you want to run as a command line argument like so:
+The easiest way is to just provide the name of the tests you want to run as a `-F` command line argument:
 
 ```shell
-./run-local.sh "description of the test to run"
+./run-local.sh -F "description of the test to run"
 ```
 
 The argument is passed as a regular expression that will match all `Describe`, `Context` or `It` closure descriptions in the suite.
 So, e.g. if you want to run all tests that use mTLS, you can run:
 
 ```shell
-./run-local.sh mTLS
+./run-local.sh -F mTLS
 ```
 
 However, if you want to run exactly one specific test, make sure you pass the exact description of the matching `It` closure:
 
 ```shell
-./run-local.sh "Correctly terminates mTLS requests"
+./run-local.sh -F "Correctly terminates mTLS requests"
 ```
 
-Alternatively, you add an `F` to your `Describe`, `Context` or `It` closures.
-Don't forget to do a local commit before running the tests, else BOSH will fail to produce a release. The git repo must be clean before running the tests.
+Running tests in focus will also preserve bosh container running after tests complete, so that you can easily run tests again without having to wait for bosh set-up again.
+
+### Bosh persistence
+
+Because BOSH setup takes a while (it starts from scratch with bosh create-env), it is useful to preserve container with bosh already configured to run tests. This can be done either by providing test focus as described above, or `-k` (keep) switch to `run-local.sh` and `run-shell.sh` scripts, e.g. `run-shell.sh -k`. Once initial setup is complete, scripts will output message about how to get back into running container:
+
+```text
+KEEP_RUNNING is true and bosh remains running.
+Re-enter container via: docker exec -it b7c767c5c0e4 bash
+
+Stop with: docker stop b7c767c5c0e4
+```
+
+After you have completed your work and have stopped the container, it is required you do further cleanup. In order to have working overlay2 filesystem for docker-cpi, it is necessary to mount ext4 based storage from `/workspace/docker-in` into bosh container. Each container gets its own temporary space, which in case of containers that keep running is not deleted when the tests complete. The scripts will output location of this temporary storage:
+
+```text
+*** KEEP_RUNNING enabled. Please clean up docker scratch after removing containers: /workspace/docker-in/scratch-19517
+```
--- a/acceptance-tests/README.md
+++ b/acceptance-tests/README.md
@@ -14,7 +14,7 @@ cd acceptance-tests
 
 ### Running on Docker for Mac
 
-Unfortunately rosetta on arm based macs doesn't support docker in docker required for bosh docker-cpi used in this test setup. You will either have to run on x86 based mac or some remote x86 workstation. Virtualization using QUEMU is possible, but so slow the tests fail on timeouts.
+Unfortunately rosetta on arm based macs doesn't support docker in docker required for bosh docker-cpi used in this test setup. You will either have to run on x86 based mac or some remote x86 workstation. Virtualization using QEMU is possible, but so slow the tests fail on timeouts.
 
 The BOSH Docker CPI requires cgroups v1 to be active. Docker for Mac since 4.3.x uses cgroups v2 by default.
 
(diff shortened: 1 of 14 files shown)


acceptance-tests/README.md
```diff
@@ -19,6 +14,8 @@ cd acceptance-tests
 
 ### Running on Docker for Mac
 
+Unfortunately rosetta on arm based macs doesn't support docker in docker required for bosh docker-cpi used in this test setup. You will either have to run on x86 based mac or some remote x86 workstation. Virtualization using QUEMU is possible, but so slow the tests fail on timeouts.
```
peanball, line 17:
```suggestion
Unfortunately rosetta on arm based macs doesn't support docker in docker required for bosh docker-cpi used in this test setup. You will either have to run on x86 based mac or some remote x86 workstation. Virtualization using QEMU is possible, but so slow the tests fail on timeouts.
```

acceptance-tests/README.md
```diff
@@ -82,24 +79,40 @@ The output at the end should be:
...
 However, if you want to run exactly one specific test, make sure you pass the exact description of the matching `It` closure:
 
 ```shell
-./run-local.sh "Correctly terminates mTLS requests"
+./run-local.sh -F "Correctly terminates mTLS requests"
 ```
 
-Alternatively, you add an `F` to your `Describe`, `Context` or `It` closures.
-Don't forget to do a local commit before running the tests, else BOSH will fail to produce a release. The git repo must be clean before running the tests.
+Running tests in focus will also preserve bosh container running after tests complete, so that you can easily run tests again without having to wait for bosh set-up again.
```
peanball, line 101:
```suggestion
Running tests in focus will also preserve the bosh container running after tests complete, so that you can easily run tests again without having to wait for bosh set-up again.
```

acceptance-tests/README.md
```diff
@@ -82,24 +79,40 @@ The output at the end should be:
...
+
+### Bosh persistence
```
peanball, line 103:
```suggestion
### Persistent BOSH
```

acceptance-tests/README.md
```diff
@@ -82,24 +79,40 @@ The output at the end should be:
...
+
+Because BOSH setup takes a while (it starts from scratch with bosh create-env), it is useful to preserve container with bosh already configured to run tests. This can be done either by providing test focus as described above, or `-k` (keep) switch to `run-local.sh` and `run-shell.sh` scripts, e.g. `run-shell.sh -k`. Once initial setup is complete, scripts will output message about how to get back into running container:
```
peanball, line 105:
```suggestion
Because BOSH setup takes a while (it starts from scratch with bosh create-env), it is useful to preserve the container with bosh already configured to run tests. This can be done either by providing test focus as described above, or `-k` (keep) switch to `run-local.sh` and `run-shell.sh` scripts, e.g. `run-shell.sh -k`. Once initial setup is complete, scripts will output a message about how to get back into the running container:
```

acceptance-tests/README.md
```diff
@@ -82,24 +79,40 @@ The output at the end should be:
...
+
+```text
+KEEP_RUNNING is true and bosh remains running.
+Re-enter container via: docker exec -it b7c767c5c0e4 bash
+
+Stop with: docker stop b7c767c5c0e4
+```
+
+After you have completed your work and have stopped the container, it is required you do further cleanup. In order to have working overlay2 filesystem for docker-cpi, it is necessary to mount ext4 based storage from `/workspace/docker-in` into bosh container. Each container gets its own temporary space, which in case of containers that keep running is not deleted when the tests complete. The scripts will output location of this temporary storage:
```
peanball, line 114:
```suggestion
After you have completed your work and have stopped the container, it is required you do further cleanup. In order to have a working overlay2 filesystem for docker-cpi, it is necessary to mount ext4 based storage from `/workspace/docker-in` into the bosh container. Each container gets its own temporary space, which in case of containers that keep running is not deleted when the tests complete. The scripts will output the location of this temporary storage:
```

acceptance-tests/run-local.sh
```diff
@@ -76,14 +85,17 @@ if [ "$(uname)" == "Darwin" ]; then
...
+if [ -n "$KEEP_RUNNING" ] ; then
+  echo
+  echo "*** KEEP_RUNNING enabled. Please clean up docker scratch after removing containers: ${DOCKER_SCRATCH}"
+  echo
+  docker run --privileged -v "$REPO_DIR":/repo -v "${DOCKER_SCRATCH}":/scratch/docker -e REPO_ROOT=/repo -e FOCUS="$FOCUS" -e KEEP_RUNNING="${KEEP_RUNNING}" haproxy-boshrelease-testflight bash -c "cd /repo/ci/scripts && ./acceptance-tests ; sleep infinity"
 else
-  docker run --rm --privileged -v "$REPO_DIR":/repo -e REPO_ROOT=/repo haproxy-boshrelease-testflight bash -c "cd /repo/ci/scripts && ./acceptance-tests"
+  docker run --rm --privileged -v "$REPO_DIR":/repo -v "${DOCKER_SCRATCH}":/scratch/docker -e REPO_ROOT=/repo -e KEEP_RUNNING="" haproxy-boshrelease-testflight bash -c "cd /repo/ci/scripts && ./acceptance-tests"
+  echo "Cleaning up docker scratch: ${DOCKER_SCRATCH}"
+  sudo rm -rf "${DOCKER_SCRATCH}"
```
peanball, line 100: would it be possible to do this with an `EXIT` trap in the other case?

peanball (commented):
Added some articles to the text, will try out on the `dev-box`. But it looks very promising. I like the name `KEEP_RUNNING` and that it auto-enables when you run a test in focus mode.

acceptance-tests/run-local.sh
mtekel: I think that already has `service docker stop` mapped. It is possible to map multiple things to a trap, but it gets complicated. So far this works.

peanball: My comment was (somewhat intentional, but confusingly) on the wrong line. You remove the scratch folder here for the "non-KEEP_RUNNING" part. But for the `KEEP_RUNNING` it's just a note in the README that might be overlooked.

mtekel: It also prints message in terminal. I have now also adjusted cleanup script to remove any orphaned scratch older than 2w.

peanball: great!

ci/scripts/functions-ci.sh
```diff
@@ -0,0 +1,110 @@
...
+    # If we get here, all files are skipped or in skipped dirs.
+    return 0
+}
+
+function stop_docker() {
+  echo "----- stopping docker"
+  service docker stop
+}
+
+# Build acceptance test image if not found.
```
peanball, line 34: For CI this does not apply as we're pulling the image from the registry, correct?

ci/scripts/functions-ci.sh
```diff
@@ -0,0 +1,110 @@
...
+
+    echo "----- Uploading Jammy stemcell"
+    bosh -n upload-stemcell $stemcell_jammy_path
+
+    echo "----- Uploading Bionic stemcell"
+    bosh -n upload-stemcell $stemcell_bionic_path
+
+    echo "----- Uploading os-conf (used for tests only)"
+    bosh -n upload-release --sha1 386293038ae3d00813eaa475b4acf63f8da226ef \
+    https://bosh.io/d/github.com/cloudfoundry/os-conf-release?v=22.1.2
```
peanball, line 76: while we're avoiding pinning, etc. can we consider removing the version from os-conf-release?

ci/scripts/start-bosh.sh
```diff
@@ -187,6 +186,7 @@ function main() {
     docker network create -d bridge --subnet=10.245.0.0/16 director_network
   fi
 
+  compilation_ops="`pwd`/ci/compilation.yml"
```
peanball, line 189:
```suggestion
  compilation_ops="$(pwd)/ci/compilation.yml"
```

alternatively you could also just use `$PWD`, or $REPO_ROOT, or determine the script's current path and put the other path relative to it in case someone starts `start-bosh.sh` manually...

peanball (approved):
some very minor comments still left. Everything should work as is though.

ci/scripts/functions-ci.sh
mtekel: yes. The CI job actually starts already in the haproxy-boshrelease image.

ci/scripts/functions-ci.sh
mtekel: This was added 4 years ago, to configure ssh for gingko user. I am not sure how much work it is to adjust tests to use 'bosh ssh' instead. Then this could be removed completely.

As for pinning - this is more of a helper function, so I guess it's more important that it keeps working than to get new functionality and security patches. 

We can bump it or leave unbound, but it is more of adjustment to the acceptance tests themselves, than organisation around them, which is what this PR is doing.

peanball: Agreed, out of scope.

ci/scripts/functions-ci.sh
peanball: perfect.

ci/scripts/update-manifest
```diff
@@ -5,13 +5,13 @@ RELEASE_NAME=$2
...
-: ${MANIFEST_PATH:?USAGE: ./ci/scripts/update-manifest GITHUB_OWNER RELEASE_NAME VERSION SHA1 MANIFEST_PATH}
+: "${MANIFEST_PATH:?USAGE: ./ci/scripts/update-manifest GITHUB_OWNER RELEASE_NAME VERSION SHA1 MANIFEST_PATH}"
 
 set -e -u
 
-manifest_len=$(wc -l $MANIFEST_PATH | awk '{print $1}')
-manifest_head=$(head -n `expr $manifest_len - 4` $MANIFEST_PATH)
-cat > $MANIFEST_PATH <<YAML
+manifest_len=$(wc -l "$MANIFEST_PATH" | awk '{print $1}')
+manifest_head=$(head -n "$(( manifest_len - 4))" "$MANIFEST_PATH")
```
peanball, line 13: just to get my head around this... it will trim the last 4 lines off of manifest_path?

head accepts negative counts, which just deduct from the end:
```shell
echo -e "1\n2\n3\n4\n5\n6" | head -n -4
1
2
```

peanball (approved):
very minor nit about the (pre-existing) use of a weird manifest trimming thing by counting lines, instead of `head -n -4` to cut the _last_ 4 lines.

ci/scripts/update-manifest
mtekel: Good point, fixed now.

PR #795, opened 2025-05-12
fix: properly check if TCP should be drained

maxmoehl:
When there is no HTTP health-check for a TCP frontend the drain logic doesn't work as the frontend it tries to disable does not exist. This commit makes the statement to disable the frontend during the draining depend on whether the HTTP health-heck for that TCP frontend is enbaled and if not excludes the drain logic for it.

Resolves: https://github.com/cloudfoundry/haproxy-boshrelease/issues/788

### DIFF
--- a/jobs/haproxy/templates/drain.erb
+++ b/jobs/haproxy/templates/drain.erb
@@ -63,7 +63,9 @@ echo "disable frontend health_check_http_url_proxy_protocol" | /usr/local/bin/so
     }
   end -%>
 <%- tcp.each do |tcp_proxy| -%>
+<%- unless tcp_proxy.fetch("health_check_http", nil).nil? -%>
 echo "disable frontend health_check_http_tcp-<%= tcp_proxy["name"] %>" | /usr/local/bin/socat stdio unix-connect:${sockfile}
+<%- end -%>
 <%- end -%>
 
 <%- if p("ha_proxy.enable_health_check_http") || tcp.size > 0 -%>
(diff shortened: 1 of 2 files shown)


jobs/haproxy/templates/drain.erb
```diff
@@ -63,8 +63,10 @@ echo "disable frontend health_check_http_url_proxy_protocol" | /usr/local/bin/so
     }
   end -%>
 <%- tcp.each do |tcp_proxy| -%>
+<%- if tcp_proxy.has_key?("health_check_http") -%>
```
mike-jc, line 66: What if the key exists but its value is empty?

maxmoehl: Damn you are right, the key always exists because of this:

https://github.com/cloudfoundry/haproxy-boshrelease/blob/ced08b6df5f52178ac73a93e5e4f0ff0ddc33bf8/jobs/haproxy/templates/drain.erb#L62

But if I start checking for `nil` this test starts to fail:

https://github.com/cloudfoundry/haproxy-boshrelease/blob/ced08b6df5f52178ac73a93e5e4f0ff0ddc33bf8/spec/haproxy/templates/drain_spec.rb#L63-L77

Because it hinges on the broken behavior...

maxmoehl: Figured it out, this makes the test pass again by setting it up properly:

https://github.com/cloudfoundry/haproxy-boshrelease/pull/795/files#diff-2932f7857a0fe2f21770cad3967b764fbd45a9ae5c59337ef26bd9727ae1323eR12-R14

mike-jc:
LGTM

b1tamara (approved):

PR #806, opened 2025-07-18
feat: Environment variable override for HAProxy Wrapper

peanball:
The HAProxy Wrapper starts the `haproxy` process.

Setting the config property `haproxy.env_override` allows declaring ad-hoc environment variables for the `haproxy` process call.

This allows overriding environment variables for testing and experimentation.

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -100,6 +100,12 @@ properties:
       Example
         curl -H "Host: bob.com" https://alice.com    <-- This will result in a 421 Misdirected Request
     default: false
+  ha_proxy.env_override:
+    description: |
+      Environment variables to be set ad-hoc when the HAProxy process is started. This can be used to override various settings for HAProxy and Linux process level.
+    example:
+       env_override: "LD_PRELOAD=/var/vcap/packages/haproxy/lib/libjemalloc.so.2"
+    default: ~
   ha_proxy.ssl_pem:
     description: |
       Array of private keys and certificates used for TLS handshakes with downstream clients. Each element in the array is an object containing fields 'cert_chain' and 'private_key',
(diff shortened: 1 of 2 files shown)


jobs/haproxy/spec
```diff
@@ -100,6 +100,12 @@ properties:
       Example
         curl -H "Host: bob.com" https://alice.com    <-- This will result in a 421 Misdirected Request
     default: false
+  ha_proxy.env_override:
```
maxmoehl, line 103: Why not:

```suggestion
  ha_proxy.env:
```

as we are not necessarily overwriting something.

jobs/haproxy/templates/haproxy_wrapper.erb
```diff
@@ -8,6 +8,8 @@ CONFIG=/var/vcap/jobs/haproxy/config/haproxy.config
 PID_FILE=/var/vcap/sys/run/haproxy/haproxy.pid
 DRAIN_LOCK=/var/vcap/sys/run/haproxy/drain.lock
 
+ENV_OVERRIDE=<$= p("ha_proxy.env_override", "") %>
```
maxmoehl, line 11: Would this work if you are providing multiple env vars? It would be safer to dump the env to a file and source it before launching HAProxy.

peanball: That's fair. Ideally the env is only applicable to the launched HAProxy process. The intended use case was a `LD_PRELOAD` variable. Maybe I can make `haproxy.env` a list and process / concatenation properly.

Adding a subshell around this (the only option I see) that would limit the scope of the env variables would be a bit overkill I think.

peanball: Just tried putting multiple vars into a single BOSH variable and that does not work.

maxmoehl: What about:
```
env $(grep -v '^#\|^$' ./env | tr '\n' ' ') haproxy ...
```
Possibly dropping the comment / empty line check as we can just forbid that.

peanball: fair. `env $VARS -S` to make sure the params to haproxy are passed, but that could work. I will try these days.

PR #894, opened 2026-04-13
Use the correct python binary from venv

mike-jc:
Now python requirements are installed in python environment  so we should use the right interpreter which has the installed libraries

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -330,7 +330,7 @@ jobs:
               platform: linux
               run:
                 dir: git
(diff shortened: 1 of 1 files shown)


hoffmaen (approved):

maxmoehl:
Please adjust the commit message to something which explains the change you are making and doesn't reference tickets that are not publicly accessible in its title.

mike-jc:
> Please adjust the commit message to something which explains the change you are making and doesn't reference tickets that are not publicly accessible in its title.

Done

PR #898, opened 2026-04-20
Revert HAProxy to 3.2.13

hoffmaen:
Due to a bug in HAProxy 3.2.14 and 3.2.15 (https://github.com/haproxy/haproxy/issues/3337) that results in frequent crashes  of HAProxy, an haproxy-boshrelease with the latest bosh-release changes and an unaffected HAProxy version is required.


a18e (approved):
Approval for CI

plowin (approved):
LGTM, exceptional patch-version revert to speed up segfaults with h/2 while keeping recent features like tcp-based blocking

PR #899, opened 2026-04-20
Rate limiting via socket

Dariquest:
Rate limiting can be configured via socket, so that it is possible to adjust rate limits when haproxy is under high load.


a18e (approved):
Approval for CI - did not review yet.

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -566,11 +570,7 @@ frontend https-in
     tcp-request connection reject if layer4_block
   <%- if_p("ha_proxy.connections_rate_limit.table_size", "ha_proxy.connections_rate_limit.window_size") do -%>
     tcp-request connection track-sc0 src table st_tcp_conn_rate
-    <%- if_p("ha_proxy.connections_rate_limit.block", "ha_proxy.connections_rate_limit.connections") do |block, connections| -%>
-      <%-if block -%>
-    tcp-request connection reject if { sc_conn_rate(0) gt <%= connections %> }
-      <%- end -%>
-    <%- end -%>
+    tcp-request connection reject if { var(proc.conn_rate_limit_enabled) -m bool } { sc_conn_rate(0),sub(proc.conn_rate_limit) gt 0 }
```
a18e, line 573: Why do we substract instead of comparing directly? Did you test that this is more performant?

acceptance-tests/rate_limit_test.go
```diff
@@ -165,6 +175,246 @@ var _ = Describe("Rate-Limiting", func() {
...
+		}
+		Expect(firstFailure).To(Equal(connLimit))
+		Expect(successfulRequestCount).To(Equal(connLimit))
+
+		By("Clearing stick table before overriding limit")
+		runHAProxySocketCommand(haproxyInfo, "clear table st_tcp_conn_rate")
+
+		By("Overriding the limit at runtime via socket to a higher value")
+		newLimit := connLimit * 3
+		runHAProxySocketCommand(haproxyInfo, fmt.Sprintf("experimental-mode on; set var proc.conn_rate_limit int(%d)", newLimit))
```
a18e, line 243: Did you investigate what the drawbacks of experimental-mode are? Do we even still require this? How long will this stay experimental?

Dariquest: set var is only available in experimental mode (since HAProxy 2.4, HAProxy Enterprise 2.4r1). Unfortunately, this operation is still in development and may change in the future (see [HAProxy documentation](https://www.haproxy.com/documentation/haproxy-runtime-api/reference/set-var/)).

jobs/haproxy/templates/haproxy.config.erb
Mrizwanshaik: Yeah I tried with comparison first but turns our haproxy comparator expects static integer on the right, not variables reference in any form. Subtraction approach is a way to compare against dynamic variables.

acceptance-tests/rate_limit_test.go
Mrizwanshaik: This is by design from haproxy version until now, this commands are still in development mode and may change later on. see: https://www.haproxy.com/documentation/haproxy-runtime-api/reference/set-var/ and https://www.haproxy.com/documentation/haproxy-runtime-api/reference/experimental-mode/


Drawback: Its per-connection only - doesn't persist across sessions. experimental-mode on is required because set var is classified as an experimental/unstable feature in HAProxy's runtime API

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -432,11 +440,7 @@ frontend http-in
...
```
a18e, line 443:
```suggestion
    tcp-request connection silent-drop if { var(proc.conn_rate_limit_enabled) -m bool } { sc_conn_rate(0),sub(proc.conn_rate_limit) gt 0 }
```
I think we should do this, even though it's a breaking change.
Yes, this will be worse for valid clients in but we work under the assumption that rate limits will only be enforced under heavy load scenarios where it's acceptable to negatively affect some valid traffic in order to prevent a complete outage.

Mrizwanshaik:
Decided to not go this way during CFN sprint review (29.04.2026), therefore move this to draft.

PR #901, opened 2026-04-21
fix: Prometheus Metrics endpoint is always enabled because ha_proxy.stats_promex_enable is not checked properly

### LINKED ISSUE #922: ha_proxy.stats_promex_enable is not evaluated properly
### Current behaviour

The Prometheus metrics endpoint is active everywhere. The value of property `ha_proxy.stats_promex_enable` is not evaluated properly:
`if_p` yields whenever a property has any value, including spec defaults
`ha_proxy.stats_promex_enable` has a default `false` in the spec, so `if_p` was always yielding, and the prometheus-exporter line was always written to the rendered config, even when the operator never set the property.

Introduced in https://github.com/cloudfoundry/haproxy-boshrelease/pull/536 

### Expected behaviour

* The boolean value is evaluated
* If the property is not set or is explicitly set to false, the Prometheus metrics endpoint should not be enabled

a18e:
- `if_p` yields whenever a property has any value, including spec defaults — it does not check the boolean value
- `ha_proxy.stats_promex_enable` has `default: false` in the spec, so `if_p` was always yielding and the `prometheus-exporter` line was always written to the rendered config, even when the property was never set by the operator
- Fix: replace `if_p(...)` with `if p(...)` so the boolean is actually evaluated

This has been the case since the feature was introduced in https://github.com/cloudfoundry/haproxy-boshrelease/pull/536 so the prometheus metrics endpoint is currently active everywhere. Merging this PR would fix this but potentially break scenarios where people rely on it but haven't set the flag to `true`.

Resolves: https://github.com/cloudfoundry/haproxy-boshrelease/issues/922

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -362,7 +362,7 @@ listen stats
     acl private src <%= p("ha_proxy.trusted_stats_cidrs") %>
     http-request deny unless private
     mode http
-  <%- if_p("ha_proxy.stats_promex_enable") do -%>
+  <%- if p("ha_proxy.stats_promex_enable") -%>
     http-request use-service prometheus-exporter if { path <%= p("ha_proxy.stats_promex_path") %> }
   <%- end -%>
     stats enable
--- a/spec/haproxy/templates/haproxy_config/stats_listener_spec.rb
+++ b/spec/haproxy/templates/haproxy_config/stats_listener_spec.rb
@@ -44,6 +44,12 @@ describe 'config/haproxy.config stats listener' do
       end
     end
 
+    context 'when ha_proxy.stats_promex_enable is false (default)' do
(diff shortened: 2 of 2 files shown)


hoffmaen (approved):
Have you checked if the wrong use of `if_p` is also present in other places?

hoffmaen (commented):
I've been poking around the code a bit, there is one more boolean property that suffers from the same bug: [reload_idle_close_on_response](https://github.com/cloudfoundry/haproxy-boshrelease/blob/18dbc0ea314829b85346c1094498886a66957127/jobs/haproxy/templates/haproxy.config.erb#L332). 

Since it defaults to true, setting it to false still enters the if_p block and enables idle-close-on-response. Can you fix that too with this PR?

b1tamara:
@hoffmaen lets keep the PR only for stats_promex

b1tamara (approved):
LGTM

PR #904, opened 2026-04-23
Bind health-check listeners on `binding_ip` and `v4v6`

s4heid:
The four health-check listeners (`health_check_http_url`, its proxy protocol variant, and the per-tcp-proxy `health_check_http_tcp-<name>` pair) previously bound on `:<port>`, which is IPv4-wildcard only. On IPv6-only deployments (`binding_ip: "::"`) the listeners were unreachable, causing monit host checks against `localhost` (IPv4) to fail and restart HAProxy.

Bind them on `ha_proxy.binding_ip` and append `v4v6` like the other frontends so dual-stack works when `ha_proxy.v4v6` is enabled.

Behavior change: when `binding_ip` is set to a specific IPv4 address, the health-check listeners now bind only on that address instead of all IPv4 interfaces. Operators probing health via a different NIC must update their probe targets accordingly.

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -378,7 +378,7 @@ listen stats
 
 <% if p("ha_proxy.enable_health_check_http") %>
 listen health_check_http_url
-    bind :<%= p("ha_proxy.health_check_port") %>
+    bind <%= p("ha_proxy.binding_ip") %>:<%= p("ha_proxy.health_check_port") %> <%= v4v6 %>
     mode http
     option httpclose
     monitor-uri /health
@@ -391,7 +391,7 @@ listen health_check_http_url
 
 <%- if enable_additional_health_check_proxy -%>
 listen health_check_http_url_proxy_protocol
-    bind :<%= p("ha_proxy.health_check_port") + 1 %> accept-proxy
+    bind <%= p("ha_proxy.binding_ip") %>:<%= p("ha_proxy.health_check_port") + 1 %> accept-proxy <%= v4v6 %>
     mode http
     option httpclose
     monitor-uri /health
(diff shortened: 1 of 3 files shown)


jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -1111,7 +1111,7 @@ listen health_check_http_tcp-<%= tcp_proxy["name"] %>
 
     <%- if enable_additional_health_check_proxy -%>
 listen health_check_http_tcp-<%= tcp_proxy["name"] %>_proxy_protocol
-    bind :<%= tcp_proxy["health_check_http"] + 1 %> accept-proxy
+    bind <%= p("ha_proxy.binding_ip") %>:<%= tcp_proxy["health_check_http"] + 1 %> accept-proxy <%= v4v6 %>
```
peanball, line 1114: ~~v4v6 somewhat negates the binding IP I think. It's only used with the default address (i.e. when it's not explicitly given). Binding to `1.2.3.4:8080 v4v6` will not magically bind an IPv6 socket if I understand the HAProxy docs correctly.~~

~~Can you test whether those settings might be mutually exclusive on a real setup, or rather if they have the effect you desire?~~

Nevermind, the `v4v6` flag is only set if the binding ip is `::` (i.e. IPv6 default address) and the `v4v6` config property is set:
https://github.com/cloudfoundry/haproxy-boshrelease/blob/f072fbf6e89579b2f29e725ad249f89a46ae5411/jobs/haproxy/templates/haproxy.config.erb#L171-L178

This is just an extension of what we have in other locations that was just never applied to health checks.

peanball (dismissed):
Looks great. Extends the logic for `v4v6` to the health check endpoints, where it was so far missing.

Approving for CI.

b1tamara (approved):
Approving to run CI

peanball (approved):
approving for CI, hopefully one final time :)

PR #924, opened 2026-05-27
fix: pinned docker-cpi version

mike-jc:
## Summary

1. Pinned the Docker-CPI image to the workable version (Bosh v282.1.13).
- This image is currently actively developed, and periodically, some bugs and/or breaking changes are introduced, causing the acceptance tests to fail. The bugs are fixed in future versions, but in the meantime, we are blocked.
- Previously, we always used the image with the tag `latest`. Unfortunately, there are no release-like tags with stable versions, so I have chosen the latest version that passed the tests.
2. Changed resource type for Docker CPI to `registry-image`.
- The resource type `docker-image` does not support pulling images by digest. 
3. Added the missed triggers to automatically rerun tests when the haproxy-boshrelease image changes.


mike-jc:
Closed in favor of #929

PR #927, opened 2026-05-28
fix: Support tcp connection rejection & rate-limiting when Proxy protocol is enabled

b1tamara:
Haproxy connection rate limiting applies a TCP reject of the connection whenever the limit is exceeded.

When the proxy protocol is enabled, the source IP during connection initiation is the LB's IP address. The proxy protocol header is only conveyed once the session is established. 

The current PR provides a fix to determine whether TCP reject of the connection or TCP reject of the session will be applied, depending on the accept proxy property.

### DIFF
--- a/acceptance-tests/access_control_test.go
+++ b/acceptance-tests/access_control_test.go
@@ -140,4 +140,42 @@ var _ = Describe("Access Control", func() {
 		By("Allowing TCP connections from non-blocklisted CIDRs (request from 127.0.0.1 on HAProxy VM)")
 		expectTestServer200(http.Get("http://127.0.0.1:11000"))
 	})
+
+	It("Rejects IPs in TCP-layer blocklisted CIDRs when proxy protocol is enabled", func() {
+		haproxyBackendPort := 12000
+
+		opsfileTCPBlocklistWithProxyProtocol := opsfileTCPBlocklist + `
+# Enable Proxy Protocol
+- type: replace
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/accept_proxy?
+  value: true
+`
+		haproxyInfo, _ := deployHAProxy(baseManifestVars{
+			haproxyBackendPort:    haproxyBackendPort,
+			haproxyBackendServers: []string{"127.0.0.1"},
+			deploymentName:        deploymentNameForTestNode(),
+		}, []string{opsfileTCPBlocklistWithProxyProtocol}, map[string]interface{}{
+			// traffic from test runner appears to come from this CIDR block
+			"cidr_blocklist_tcp": []string{"10.1.1.1/32"},
+		}, true)
+
+		closeLocalServer, localPort := startDefaultTestServer()
+		defer closeLocalServer()
+
+		closeBackendTunnel := setupTunnelFromHaproxyToTestServer(haproxyInfo, haproxyBackendPort, localPort)
+		defer closeBackendTunnel()
+
+		By("Denying TCP connections from blocklisted CIDRs via Proxy Protocol header (source IP 10.1.1.1)")
+		// Send a request with a Proxy Protocol header advertising a source IP in the blocked 10.0.0.0/8 CIDR.
+		// With accept_proxy enabled, HAProxy parses the PROXY header and uses the advertised source IP
+		// for the tcp-request session phase ACL evaluation — so this must be rejected.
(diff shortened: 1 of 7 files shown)


a18e (dismissed):
LGTM

acceptance-tests/access_control_test.go
```diff
@@ -140,4 +140,42 @@ var _ = Describe("Access Control", func() {
...
 	})
+
+	It("Rejects IPs in TCP-layer blocklisted CIDRs when proxy protocol is enabled", func() {
+		haproxyBackendPort := 12000
+
+		opsfileTCPBlocklistWithProxyProtocol := opsfileTCPBlocklist + `
+# Enable Proxy Protocol
+- type: replace
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/accept_proxy?
+  value: true
```
a18e, line 151: That's a bit weird to read, maybe just add the whole content of the `opsfileTCPBlocklist` again?

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
```diff
@@ -585,6 +585,41 @@
...
+        let(:properties) do
+          default_properties.merge({ 'accept_proxy' => true, 'cidr_blocklist_tcp' => [] })
+        end
+
+        it 'still includes the layer4_block acl and reject rule (always present)' do
+          expect(frontend_https).to include('acl layer4_block src -f /var/vcap/jobs/haproxy/config/blocklist_cidrs_tcp.txt')
+          expect(frontend_https).to include('tcp-request session reject if layer4_block')
+        end
+      end
+    end
```
a18e, line 622: so for the http frontend (file above), you're testing:
* accept_proxy' => true, 'cidr_blocklist_tcp' => ['172.168.4.1/32', '10.2.0.0/16']
*  'accept_proxy' => true 
* 'cidr_blocklist_tcp' => []

First two have the same outcome/ `it` block. Third one is different (connection instead of session)

Here, you're testing
* 'accept_proxy' => true, 'cidr_blocklist_tcp' => ['172.168.4.1/32', '10.2.0.0/16']
* 'accept_proxy' => true
* 'accept_proxy' => true, 'cidr_blocklist_tcp' => []

All three have the same `it` block.
This is a bit inconsistent.
Also I'm not sure if we need to test that extensively, IMO it's enough to test the following for both frontends:
* accept_proxy' => true, 'cidr_blocklist_tcp' => ['172.168.4.1/32', '10.2.0.0/16']
* 'cidr_blocklist_tcp' => []

a18e (commented):
Minor remarks

acceptance-tests/access_control_test.go
```diff
@@ -140,4 +140,42 @@ var _ = Describe("Access Control", func() {
...
+
+		closeBackendTunnel := setupTunnelFromHaproxyToTestServer(haproxyInfo, haproxyBackendPort, localPort)
+		defer closeBackendTunnel()
+
+		By("Denying TCP connections from blocklisted CIDRs via Proxy Protocol header (source IP 10.1.1.1)")
+		// Send a request with a Proxy Protocol header advertising a source IP in the blocked 10.0.0.0/8 CIDR.
+		// With accept_proxy enabled, HAProxy parses the PROXY header and uses the advertised source IP
+		// for the tcp-request session phase ACL evaluation — so this must be rejected.
+		err := performProxyProtocolRequestWithSourceIP(haproxyInfo.PublicIP, 80, "/", "10.1.1.1")
+		Expect(err).To(HaveOccurred())
```
a18e, line 173: blocking this request may not create an error, this could explain the acceptance test failures.

https://concourse.arp.cloudfoundry.org/teams/main/pipelines/haproxy-boshrelease/jobs/acceptance-tests-pr/builds/287#L69d7d983:7337:7340

Other example:
https://github.com/cloudfoundry/haproxy-boshrelease/blob/64efa7a5df589a9e0ce551fcc2166a121c9c0fa1/acceptance-tests/proxy_protocol_test.go#L113

b1tamara: I added another check for syscall.ECONNRESET

a18e (approved):

spec/haproxy/templates/haproxy_config/frontend_https_spec.rb
b1tamara: You are right. Adapted the fronend_http tests

PR #930, opened 2026-06-01
feat: Support runtime control of connection rate limiting via socket

b1tamara:
The PR introduces process-level variables (for the properties block and connections) for the connection rate limit feature, enabling runtime control via the HAProxy Runtime API. It allows dynamic start/stop of rate limiting and on-the-fly threshold changes
Co-Authored-By: @Dariquest and @Mrizwanshaik

### DIFF
--- a/docs/rate_limiting.md
+++ b/docs/rate_limiting.md
@@ -134,7 +134,7 @@ echo "show table st_tcp_conn_rate" | socat stdio /var/vcap/sys/run/haproxy/stats
 
 The `connections_rate_limit.block` flag and `connections_rate_limit.connections` threshold are stored as HAProxy process-level variables (`proc.conn_rate_block` and `proc.conn_rate_limit`) and can be changed at runtime without a reload. This requires `ha_proxy.master_cli_enable: true` or `ha_proxy.stats_enable: true`.
 
-The socket is located at `/var/vcap/sys/run/haproxy/stats.sock`. You will likely need `sudo` to access it.
+The socket is located at `/var/vcap/sys/run/haproxy/stats.sock`. You will likely need `root` permissions to access it.
 
 > **Note:** The `tcp-request connection reject` rule is always present in the config as long as `connections_rate_limit.table_size` and `connections_rate_limit.window_size` are set. Enforcement is controlled entirely at runtime via `proc.conn_rate_block` and `proc.conn_rate_limit`. Setting `connections_rate_limit.connections` and `connections_rate_limit.block` in the manifest only sets their **initial values** at startup — they can be freely overridden via socket without a reload.
 
--- a/docs/rate_limiting.md
+++ b/docs/rate_limiting.md
@@ -110,7 +110,7 @@ frontend http-in
 ```
 
 ## Querying Current Stick-Table Status
-To get more insight into what is going on inside HAProxy regarding its rate limits, you can query the stats socket to get the raw table data:
+To get more insight into what is going on inside HAProxy regarding its rate limits, you can query the stats socket at `/var/vcap/sys/run/haproxy/stats.sock` to get the raw table data:
 
 ```bash
 $ echo "show table st_http_req_rate" | socat /var/vcap/sys/run/haproxy/stats.sock -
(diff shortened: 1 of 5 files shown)


jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -230,6 +230,15 @@ end
     abort "Conflicting configuration: enable_redispatch works only with retries > 0"
   end
 
+  # Safety guard: block=true without connections would cause every client with >= 1 connection to be blocked (total lockout)
+  if_p("ha_proxy.connections_rate_limit.table_size", "ha_proxy.connections_rate_limit.window_size") do
+    if p("ha_proxy.connections_rate_limit.block", false)
+      if !p("ha_proxy.connections_rate_limit.connections", nil)
+        abort "Conflicting configuration: connections_rate_limit.connections must be set when connections_rate_limit.block is true (otherwise every client with >= 1 connection would be blocked)"
```
hoffmaen: The abort is reasonable as a UX guard, but the rationale ("every client with >= 1 connection would be blocked") is not what would actually happen with the new rule: when proc.conn_rate_limit is undefined, the second guard `{ var(proc.conn_rate_limit) -m int gt 0 }` evaluates falsy and the reject does not fire. So the real failure mode is "block flag honored but never enforced" (silent no-op), not "everyone locked out". The latter was true of the old reject rule format, but no longer applies. Suggest rewording, e.g.:

```suggestion
        abort "connections_rate_limit.connections must be set in the manifest as the initial threshold when block is true; otherwise rate-limiting will be silently disabled until a value is set via the runtime API."
```

docs/rate_limiting.md
```diff
@@ -115,7 +115,59 @@ To get more insight into what is going on inside HAProxy regarding its rate limi
...
+echo "show table st_tcp_conn_rate" | socat stdio /var/vcap/sys/run/haproxy/stats.sock | sort -t= -k2 -rn | head -1
 ```
 
 > Note: You will likely need `sudo` permission to run socat.
+
+## Control Connection Rate Limiting via HAProxy Runtime API
+
+The `connections_rate_limit.block` flag and `connections_rate_limit.connections` threshold are stored as HAProxy process-level variables (`proc.conn_rate_block` and `proc.conn_rate_limit`) and can be changed at runtime without a reload. This requires `ha_proxy.master_cli_enable: true` or `ha_proxy.stats_enable: true`.
+
+The socket is located at `/var/vcap/sys/run/haproxy/stats.sock`. You will likely need `sudo` to access it.
```
hoffmaen:
```suggestion
The socket is located at `/var/vcap/sys/run/haproxy/stats.sock`. You will likely need `root` permissions to access it.
```

docs/rate_limiting.md
```diff
@@ -115,7 +115,59 @@ To get more insight into what is going on inside HAProxy regarding its rate limi
...
+
+To find the IP with the highest connection rate, use:
+
+```bash
+echo "show table st_tcp_conn_rate" | socat stdio /var/vcap/sys/run/haproxy/stats.sock | sort -t= -k2 -rn | head -1
 ```
 
 > Note: You will likely need `sudo` permission to run socat.
+
+## Control Connection Rate Limiting via HAProxy Runtime API
```
hoffmaen, line 133: Maybe make the two new `proc.` variables more prominent, by introducing them in an unsorted list, similar to the variables and options listed in the section "Configuration Options". It was first a bit hard for me to understand why we have these proc variables now.

One point about the naming: 
`connections_rate_limit.block` -> `proc.conn_rate_block`
`connections_rate_limit.connections` -> `proc.conn_rate_limit`

Is inconsistent. If we could still change the it and accept it's breaking, I'd prefer the following variable naming:

`connection_rate_limit.enabled` -> `proc.connection_rate_limit_enabled`
`connection_rate_limit.connections` -> `proc.connection_rate_limit_connections`

This would be more consistent, and having `enabled` instead of `block` indicates that this is a boolean switch.

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -324,6 +333,12 @@ global
   <%- if backend_match_http_protocol && backends.length == 2 -%>
     set-var proc.h2_alpn_tag str(h2)
   <%- end -%>
+  <%- if_p("ha_proxy.connections_rate_limit.table_size", "ha_proxy.connections_rate_limit.window_size") do -%>
+    <%- if_p("ha_proxy.connections_rate_limit.connections") do |conn_rate_connections| -%>
+    set-var proc.conn_rate_limit int(<%= conn_rate_connections %>)
+    <%- end -%>
```
mike-jc, line 339: Did I understand correctly that the existence of `connections_rate_limit.connections` in manifest controls if haproxy config will have a process variable, and when it is not set in manifest, the rate-limit control is not possible on the fly? If so, it does not look straightforward to me.
Can we set the variable to zero if the parameter is not set in the manifest? Similar to what we do for proc.conn_rate_block.

```suggestion
    set-var proc.conn_rate_limit int(<%= p("ha_proxy.connections_rate_limit.connections", 0) %>)
```

b1tamara: We store the value from the manifest in a process-level variable if defined. If not, it can be set via the socket later. We cannot set the default value to 0, as 0 is a very wrong threshold.

docs/rate_limiting.md
b1tamara: Renamed variables:
connection_rate_limit.block -> proc.connection_rate_limit_block
connection_rate_limit.connections -> proc.connection_rate_limit_connections

Enhanced the documentation

hoffmaen (approved):

PR #937, opened 2026-06-22
dynamic allowlisting on host and ip prefix

mike-jc:
- HAProxy config adds request blocking logic:
   - Source IP is converted to binary
   - Two lookup keys are built: `domain|ip-version|ip-binary` for the exact hostname, and `.parent-domain|ip-version|ip-binary` for all subdomains
      - IP version is introduced to differentiate binary prefixes for IPv4 and IPv6 
   - Both keys are checked against the ACL list; if neither matches, the request is rejected with 403
   - The ACL list is loaded from the file on startup and can be updated at runtime via HAProxy socket using `add acl / del acl` commands

### DIFF
--- a/jobs/haproxy/templates/domain_ip_allowlist.acl.erb
+++ b/jobs/haproxy/templates/domain_ip_allowlist.acl.erb
@@ -1,0 +1,25 @@
+# generated from domain_ip_allowlist.acl.erb
+<%
+  require "base64"
+  require 'zlib'
+  require 'stringio'
+
+  if_p("ha_proxy.domain_ip_allowlist") do |lines|
+    uncompressed = ''
+    if lines.is_a?(Array)
+      uncompressed << "\# detected ACL entries provided as array in cleartext format\n"
+      lines.each do |line|
+        uncompressed << line << "\n"
+      end
+    else
+      gzplain = Base64.decode64(lines)
+      gz = Zlib::GzipReader.new(StringIO.new(gzplain))
+      uncompressed = gz.read
+    end
+%>
+# BEGIN domain-ip allowlist
+<%= uncompressed %>
+# END domain-ip allowlist
+<%
+  end
+%>
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -289,6 +289,28 @@ end
   # to keep backward compatibility enable_additional_health_check_proxy if expect_proxy_cidrs is not empty.
   enable_additional_health_check_proxy = p("ha_proxy.enable_additional_health_check_proxy", p("ha_proxy.expect_proxy_cidrs", []).size > 0)
 
+  domain_ip_allowlist_config = <<~TXTBLOCK
+
+# Checks the request's host header and source IP against an ACL file.
+# Each ACL entry has the format: domain|ip-version|ip-binary-prefix
+#   - A domain starting with "." matches all subdomains (e.g. ".example.com" matches "foo.example.com")
+#   - ip-binary-prefix is the IP address in binary; a shorter prefix matches a wider range (like CIDR)
+# Examples: .bar.example.com|4|010101010101     (IPv4, matches /12 prefix)
+#           foo.bar.example.com|6|010101...01   (IPv6 exact host)
+http-request set-var(req.ip_bin) src,base2
+http-request set-var(req.ip_proto) str(4)
+http-request set-var(req.ip_proto) str(6) if { var(req.ip_bin),length gt 32 }
+http-request set-var(req.domain) hdr(host),host_only,lower
+http-request set-var-fmt(req.domain_key) %[var(req.domain)]|%[var(req.ip_proto)]|%[var(req.ip_bin)]
+http-request set-var(req.parent_domain) var(req.domain),regsub(^[^\.]+,)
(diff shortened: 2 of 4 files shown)


jobs/haproxy/templates/domain_ip_allowlist.acl.erb
```diff
@@ -0,0 +1,25 @@
+# generated from domain_ip_allowlist.acl.erb
```
a18e, line 1: We usually use `.txt` as file ending (instead of `.acl`), e.g. https://github.com/cloudfoundry/haproxy-boshrelease/blob/master/jobs/haproxy/templates/whitelist_cidrs.txt.erb

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -289,6 +289,28 @@ end
...
+# Each ACL entry has the format: domain|ip-version|ip-binary-prefix
+#   - A domain starting with "." matches all subdomains (e.g. ".example.com" matches "foo.example.com")
+#   - ip-binary-prefix is the IP address in binary; a shorter prefix matches a wider range (like CIDR)
+# Examples: .bar.example.com|4|010101010101     (IPv4, matches /12 prefix)
+#           foo.bar.example.com|6|010101...01   (IPv6 exact host)
+http-request set-var(req.ip_bin) src,base2
+http-request set-var(req.ip_proto) str(4)
+http-request set-var(req.ip_proto) str(6) if { var(req.ip_bin),length gt 32 }
+http-request set-var(req.domain) hdr(host),host_only,lower
+http-request set-var-fmt(req.domain_key) %[var(req.domain)]|%[var(req.ip_proto)]|%[var(req.ip_bin)]
```
a18e, line 304: There's probably a way to inline the `ip_proto` variable here and save on variable allocation.

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -289,6 +289,28 @@ end
...
+http-request set-var(req.parent_domain) var(req.domain),regsub(^[^\.]+,)
+http-request set-var-fmt(req.parent_domain_key) %[var(req.parent_domain)]|%[var(req.ip_proto)]|%[var(req.ip_bin)]
+acl domain_ip_matched var(req.domain_key) -m beg -f opt@/var/vcap/jobs/haproxy/config/domain_ip_allowlist.acl
+acl parent_domain_ip_matched var(req.parent_domain_key) -m beg -f opt@/var/vcap/jobs/haproxy/config/domain_ip_allowlist.acl
```
a18e, line 308: Oh, it's unfortunate that we have to match again with the parent domain, but I guess there's no way around it.
However, couldn't we skip the `parent_domain_ip_matched` acl, as well as `http-request set-var-fmt(req.parent_domain_key) %[var(req.parent_domain)]|%[var(req.ip_proto)]|%[var(req.ip_bin)]` if `domain_ip_matched` is already true?

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -289,6 +289,28 @@ end
...
+
+# Checks the request's host header and source IP against an ACL file.
+# Each ACL entry has the format: domain|ip-version|ip-binary-prefix
+#   - A domain starting with "." matches all subdomains (e.g. ".example.com" matches "foo.example.com")
+#   - ip-binary-prefix is the IP address in binary; a shorter prefix matches a wider range (like CIDR)
+# Examples: .bar.example.com|4|010101010101     (IPv4, matches /12 prefix)
+#           foo.bar.example.com|6|010101...01   (IPv6 exact host)
+http-request set-var(req.ip_bin) src,base2
+http-request set-var(req.ip_proto) str(4)
+http-request set-var(req.ip_proto) str(6) if { var(req.ip_bin),length gt 32 }
```
a18e, line 302: I would be surprised if there isn't a HAProxy native way of checking if `src` is ipv6.

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -289,6 +289,28 @@ end
...
+http-request set-var(req.domain) hdr(host),host_only,lower
+http-request set-var-fmt(req.domain_key) %[var(req.domain)]|%[var(req.ip_proto)]|%[var(req.ip_bin)]
+http-request set-var(req.parent_domain) var(req.domain),regsub(^[^\.]+,)
```
a18e, line 305: > regsub invokes a full regex engine (PCRE or POSIX depending on build), which is noticeably heavier than HAProxy's plain string converters. For a fixed-structure operation like "drop the first label," you can use the field converter instead, which just does delimiter-based splitting — no regex compilation or matching involved.
> ```
> http-request set-var(req.parent_domain) req.domain,field(2,.,-1)
> ```

a18e (commented):
Absolutely fine for the current POC state.
I am however worried about the performance impact of this if we decide to use it.

Wondering if it's worth to investigate an alternative approach:
- a separate list that contains only the domains and their pre-calculated parent domains 
- First check if  `hdr(host)` is contained in this list (via full match instead of `beg`, which should be more performant) 
- Only run the other acls if this first check is true.

Could be more performant in the end since we expect that only a small fraction of all domains will be in this per-domain-allowlist.

Also added some more comment that should be tackled if/when we decide to proceed.

PR #948, opened 2026-07-10
Introduce exclude cidrs list from rate-limiting feature

Mrizwanshaik:
- Adds an optional `ha_proxy.connections_rate_limit.exclude_cidrs` property that  exempts specific source ranges (e.g. internal/NAT ranges, health checkers,trusted peers) from **connection-based** rate limiting.                                              

- Excluded sources are **still tracked** in the `st_tcp_conn_rate` stick-table (so they remain visible via `show table st_tcp_conn_rate`), but are **never  rejected**, regardless of the configured threshold or the runtime `block`                         setting.

### DIFF
--- a/acceptance-tests/rate_limit_test.go
+++ b/acceptance-tests/rate_limit_test.go
@@ -153,11 +153,11 @@ var _ = Describe("Rate-Limiting", func() {
 
 	It("Connection Based Limiting Works with Proxy Protocol enabled", func() {
 		connLimit := 5
-		opsfileAcceptProxy := fmt.Sprintf(`---
+		opsfileAcceptProxy := `---
 - type: replace
-  path: %s/accept_proxy?
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/accept_proxy?
   value: true
-`, rateLimitPropsPath)
+`
 
 		haproxyBackendPort := 12000
 		haproxyInfo, _ := deployHAProxy(baseManifestVars{
@@ -321,24 +321,22 @@ var _ = Describe("Rate-Limiting", func() {
 	})
 })
 
-const rateLimitPropsPath = "/instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy"
-
 // requestsRateLimitOps builds an opsfile fragment configuring the requests_rate_limit properties.
 func requestsRateLimitOps(requests int, windowSize, tableSize string, block bool) string {
 	return fmt.Sprintf(`---
 - type: replace
-  path: %[1]s/requests_rate_limit?/requests
-  value: %[2]d
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/requests_rate_limit?/requests
+  value: %d
 - type: replace
-  path: %[1]s/requests_rate_limit/window_size?
-  value: %[3]s
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/requests_rate_limit/window_size?
+  value: %s
 - type: replace
-  path: %[1]s/requests_rate_limit/table_size?
-  value: %[4]s
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/requests_rate_limit/table_size?
+  value: %s
 - type: replace
-  path: %[1]s/requests_rate_limit/block?
-  value: %[5]t
-`, rateLimitPropsPath, requests, windowSize, tableSize, block)
+  path: /instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy/requests_rate_limit/block?
+  value: %t
+`, requests, windowSize, tableSize, block)
 }
 
 // connectionsRateLimitOps builds an opsfile fragment configuring the connections_rate_limit properties.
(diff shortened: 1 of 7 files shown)


spec/haproxy/templates/haproxy_config/rate_limit_spec.rb
```diff
@@ -113,6 +113,34 @@
...
+      let(:properties) do
+        temp_properties.deep_merge({ 'connections_rate_limit' => { 'exclude_cidrs' => ['10.0.0.0/8', '192.168.0.0/16'] } })
+      end
+
+      it 'adds the exclusion acl to http-in and https-in frontends' do
+        expect(frontend_http).to include('acl rate_limit_exclude src -f /var/vcap/jobs/haproxy/config/rate_limit_exclusion_cidrs.txt')
+        expect(frontend_https).to include('acl rate_limit_exclude src -f /var/vcap/jobs/haproxy/config/rate_limit_exclusion_cidrs.txt')
+      end
+
+      it 'still tracks excluded sources in the stick-table' do
```
b1tamara: This test does not check whether the excluded sources are in the stick-table, but only whether the line is included in the config.

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -239,6 +239,12 @@ end
     end
   end
 
+  # Connections rate-limit exclusion: sources matching exclude_cidrs are tracked but never rejected
+  conn_rate_limit_exclude = ""
+  if_p("ha_proxy.connections_rate_limit.exclude_cidrs") do
```
b1tamara: I am thinking whether it is simpler to always have a default empty array and get rid of `if exclude_cidrs` are set. 
If the array remains empty, there are no CIDRS to exclude from rejection.

b1tamara (dismissed):
Approving to run CI

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -451,9 +457,12 @@ frontend http-in
     acl layer4_block src -f /var/vcap/jobs/haproxy/config/blocklist_cidrs_tcp.txt
     tcp-request <%= tcp_request_phase %> reject if layer4_block
   <%- if_p("ha_proxy.connections_rate_limit.table_size", "ha_proxy.connections_rate_limit.window_size") do -%>
+    <%- if_p("ha_proxy.connections_rate_limit.exclude_cidrs") do -%>
+    acl rate_limit_exclude src -f /var/vcap/jobs/haproxy/config/rate_limit_exclusion_cidrs.txt
+    <%- end -%>
     tcp-request <%= tcp_request_phase %> track-sc0 src table st_tcp_conn_rate
     # use sub() converter as variable references are only accepted as arguments to converters
-    tcp-request <%= tcp_request_phase %> reject if { var(proc.connections_rate_limit_block) -m bool } { var(proc.connections_rate_limit_connections) -m int gt 0 } { sc_conn_rate(0),sub(proc.connections_rate_limit_connections) gt 0 }
+    tcp-request <%= tcp_request_phase %> reject if { var(proc.connections_rate_limit_block) -m bool } { var(proc.connections_rate_limit_connections) -m int gt 0 } { sc_conn_rate(0),sub(proc.connections_rate_limit_connections) gt 0 }<%= conn_rate_limit_exclude %>
```
a18e: We should find out if the order of the conditions here makes a difference.
We know that >50% of the ingress traffic is landscape-internal and therefore would be matched by the exclusion list.
If we move the exclusion list to the front to be the first checked condition, will it skip the stick table fetches (which could be more expensive? If not, we need to keep this order)

jobs/haproxy/templates/haproxy.config.erb
Mrizwanshaik: changed the implementation to take empty array.

spec/haproxy/templates/haproxy_config/rate_limit_spec.rb
Mrizwanshaik: renamed the test to reflect what it does

jobs/haproxy/templates/haproxy.config.erb
Mrizwanshaik: I'd keep the current order, Reordering won't help. The stick-table work isn't in the reject rule. It's the separate track-sc0 line above it, which runs unconditionally for every  connection. By the time the reject rule runs, sc_conn_rate(0) is just a cheap in-memory read of the already-tracked entry, not a fresh lookup. So moving !rate_limit_exclude first only short-circuits that cheap read (and an src -f CIDR match is actually more expensive than the var() checks in front of it).

a18e (dismissed):
Approval for CI - did not review

hoffmaen (dismissed):
Approving for CI

acceptance-tests/rate_limit_test.go
```diff
@@ -381,3 +320,48 @@ var _ = Describe("Rate-Limiting", func() {
...
+
+// connectionsRateLimitOps builds an opsfile fragment configuring the connections_rate_limit properties.
+func connectionsRateLimitOps(connections int, windowSize, tableSize string, block bool, excludeCIDRs []string) string {
+	quotedCIDRs := make([]string, len(excludeCIDRs))
+	for i, cidr := range excludeCIDRs {
+		quotedCIDRs[i] = fmt.Sprintf("%q", cidr)
+	}
+	return fmt.Sprintf(`---
+- type: replace
+  path: %[1]s/connections_rate_limit?/connections
```
b1tamara: I would keep the path inline (as before), not as a variable. Since the OpsFile is defined once, it's okay to use the full path for better readability.

acceptance-tests/rate_limit_test.go
```diff
@@ -381,3 +320,48 @@ var _ = Describe("Rate-Limiting", func() {
...
+  value: %[2]d
+- type: replace
+  path: %[1]s/connections_rate_limit/window_size?
+  value: %[3]s
+- type: replace
+  path: %[1]s/connections_rate_limit/table_size?
+  value: %[4]s
```
b1tamara: window_size and table_size could be hard-coded values for all tests.

acceptance-tests/rate_limit_test.go
```diff
@@ -381,3 +320,48 @@ var _ = Describe("Rate-Limiting", func() {
...
+	return fmt.Sprintf(`---
+- type: replace
+  path: %[1]s/requests_rate_limit?/requests
+  value: %[2]d
+- type: replace
+  path: %[1]s/requests_rate_limit/window_size?
+  value: %[3]s
+- type: replace
+  path: %[1]s/requests_rate_limit/table_size?
+  value: %[4]s
```
b1tamara: window_size and table_size could be hard-coded values for all tests.

acceptance-tests/rate_limit_test.go
```diff
@@ -381,3 +320,48 @@ var _ = Describe("Rate-Limiting", func() {
...
 	})
 })
+
+const rateLimitPropsPath = "/instance_groups/name=haproxy/jobs/name=haproxy/properties/ha_proxy"
+
+// requestsRateLimitOps builds an opsfile fragment configuring the requests_rate_limit properties.
+func requestsRateLimitOps(requests int, windowSize, tableSize string, block bool) string {
+	return fmt.Sprintf(`---
+- type: replace
+  path: %[1]s/requests_rate_limit?/requests
```
b1tamara: Write path inline for better readability.

acceptance-tests/rate_limit_test.go
Mrizwanshaik: Different values are set for different tests, so I keep it as place holder. I am not sure if behaviour changes when hard coded?

b1tamara (approved):
Approving for CI

b1tamara:
Thank you for refactoring the rate-limiting acceptance test file to reuse the OpsFiles.

b1tamara (approved):
LGTM

